PR #2, opened 2019-04-01
Provide local accessor for `bson_t.len`

mbroadst:
In Swift 5, `bson_t` is treated as an `OpaquePointer` so we no
longer have type information for its fields. This allows us to
retrieve the `len` field by providing a custom accessor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -65,3 +65,4 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
+.vscode
```
kmahar: nit: newline at end of file

PR #4, opened 2020-05-07
SWIFT-837 Add API sketch required for SWIFT-841

kmahar:
Forgot I hadn't included this in the last PR, but this should include enough skeleton to get Neal rolling on SWIFT-841 which is adding support for reading and writing integer types and nested documents.

This is all just copying API sketch from driver/design doc. Where possible I left in existing implementations of methods from the driver.

Also I left out Codable conformance for now / removed `BSONValue`'s inheritance from `Codable`, but I'm going to open a new ticket in the epic about adding that back and copyingover conformances + BSONEncoder and BSONDecoder.

### DIFF
--- a/Sources/BSON/Document.swift
+++ b/Sources/BSON/Document.swift
@@ -50,9 +50,9 @@ public struct Document {
     /// The number of (key, value) pairs stored at the top level of this document.
     public var count: Int { fatalError("Unimplemented") }
 
-    /// A copy of the `ByteBuffer` backing this document, containing raw BSON data.
-    /// As `ByteBuffer`s implement copy-on-write, this copy will share byte storage with
-    /// this document until either the document or the returned buffer is mutated.
+    /// A copy of the `ByteBuffer` backing this document, containing raw BSON data. As `ByteBuffer`s implement
+    /// copy-on-write, this copy will share byte storage with this document until either the document or the returned
+    /// buffer is mutated.
     public var buffer: ByteBuffer { fatalError("Unimplemented") }
 
     /// Returns a `Data` containing a copy of the raw BSON data backing this document.
@@ -69,8 +69,8 @@ public struct Document {
      *  d["a"] = 1
      *  print(d["a"]) // prints 1
      *  ```
-     * A nil return value indicates that the key does not exist in the  `Document`.
-     * A true BSON null is returned as `BSON.null`.
+     * A nil return value indicates that the key does not exist in the  `Document`. A true BSON null is returned as
+     * `BSON.null`.
      */
     public subscript(key: String) -> BSON? {
         get { fatalError("Unimplemented") }
--- a/Sources/BSON/Document.swift
+++ b/Sources/BSON/Document.swift
@@ -1,0 +1,153 @@
+import Foundation
+import NIO
+
+/// A struct representing the BSON document type.
+@dynamicMemberLookup
+public struct Document {
+    /// The element type of a document: a tuple containing an individual key-value pair.
+    public typealias KeyValuePair = (key: String, value: BSON)
+
+    private var _buffer: ByteBuffer
+
+    /// An unordered set containing the keys in this document.
+    private var keySet: Set<String>
+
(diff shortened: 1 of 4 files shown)


Sources/BSON/DocumentIterator.swift
```diff
@@ -0,0 +1,57 @@
...
+    // Since a `Document` is a recursive structure, we want to enforce the use of it as a subsequence of itself,
+    // instead of something like `Slice<Document>`.
+    /// The type that is returned from methods such as `dropFirst()` and `split()`.
+    public typealias SubSequence = Document
+
+    /// Returns a `Bool` indicating whether the document is empty.
+    public var isEmpty: Bool { fatalError("Unimplemented") }
+
+    /// Returns a `DocumentIterator` over the values in this `Document`.
+    public func makeIterator() -> DocumentIterator {
```
kmahar, line 14: oh, also, I had this in the design as `some IteratorProtocol`, but apparently that only works on macOS 10.15+.... dropping < 10.15 support is probably not an option for us right now (at the very least because of Evergreen not even having 10.15 to test on, lol).
so I had to make that type public again and make this return a `DocumentIterator` 😠 

opened [SWIFT-855](https://jira.mongodb.org/browse/SWIFT-855) re macOS version support policy

Sources/BSON/Document.swift
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Initializes a new `Document` from the provided BSON data. If validate is `true` (the default), validates that
+     * the data is specification-compliant BSON.
+     *
+     * - Throws:
+     *   - `InvalidArgumentError` if the data passed is invalid BSON
+     *
+     * - SeeAlso: http://bsonspec.org/
+     */
+    public init(fromBSON bson: Data, validate: Bool = true) throws { fatalError("Unimplemented") }
```
patrickfreed, line 31: Now that I'm looking at this signature, I wonder if it would be better to have the validation separated into two separate initializers. In order for someone to call this in a non-`throws` context, they'd need to invoke it like:
```swift
try! Document(fromBSON: data, validate: false) // never fails, but requires try still
```
which is a little verbose.

kmahar: that's a good point. we can't have two identical signatures where only one throws, so I guess we could disambiguate them by using different argument labels. 
I feel like the "default" one should be validating and then we have some explicit label indicating we are not validating in the other case. something like
```swift
init(fromBSON bson: Data) throws

// only use this if you know your BSON data is already valid
init(fromValidBSON bson: Data)
```

don't love the naming here though, any ideas?

patrickfreed: another possible option is `withoutValidating:`. We could also use a static method and do `Document.fromBSONWithoutValidation(_ bson: Data)`. That said, I don't necessarily like either of these more than `fromValidBSON`.

Another question I have is whether we expect users to make use of this, or whether it's mainly for us in the driver. If it's the latter, we could internalize it and name it whatever we want.

kmahar: I'd be fine with keeping it internal and making it public later if someone asks for it.
we only have a validating version of the init right now after all, and no one has complained. we just don't do as much validation in it as we could do

patrickfreed: sgtm

kmahar: @mbroadst @nbbeeken does that sound reasonable to you?

nbbeeken: I think the validating form only as the public one sounds good. We still have to disambiguate though right? We can push the less aesthetically pleasant naming to the internal one

kmahar: ping @mbroadst, would be good to get this merged in the next couple days so we can make progress with Neal's PRs

mbroadst: SGTM

kmahar: cool I updated the initializers, I think @nbbeeken can just add the internal versions when he needs them

nbbeeken: sgtm 🚀

PR #5, opened 2020-05-14
SWIFT-841: Add support for reading/writing Integer BSON types in documents

nbbeeken:
Adds:
- DocumentIterator
- Int32, Int64, and Int
- subscripting, dynamicMember, count, keys, and values

Some manual hand testing is here, a very basic BSON Corpus test runner is there as well.
The actual corpus tests I'll hold off on committing for now to make the diff easier on the eyes.

~Currently accessing all the values for a 3 item doc sometimes takes 14 seconds 😨 that is crazy... I must've done something here that implicitly loops multiple times~


Sources/BSON/Document.swift
```diff
@@ -1,6 +1,30 @@
 import Foundation
 import NIO
 
+private let BSON_ALLOCATOR = ByteBufferAllocator()
```
kmahar, line 4: let's add a comment about this

kmahar:
regarding execution time, I wonder if maybe you could do some crude testing of where most of time is spent, maybe by e.g. inserting print statements with the current time

Sources/BSON/Document.swift
```diff
@@ -14,10 +39,44 @@ public struct Document {
...
+        let start = self._buffer.writerIndex
+
+        // reserve space for our size that will be calculated
+        self._buffer.writeInteger(0, endianness: .little, as: Int32.self)
+
+        for (key, bson) in keyValuePairs {
+            self._buffer.writeInteger(UInt8(bson.bsonValue.bsonType.rawValue), as: UInt8.self)
+            self._buffer.writeCString(key)
+            // swiftlint:disable:next force_try
+            try! bson.bsonValue.write(to: &self._buffer)
```
kmahar, line 59: I guess we are `try!`ing here since `BSONValue.write` throws. however, it looks like so far neither of our implementations actually need to throw, since `writeInteger` doesn't. maybe we can change the protocol method to not throw, and then we don't need to `try!` here?

Sources/BSON/Document.swift
```diff
@@ -14,10 +39,44 @@ public struct Document {
...
+        }
+        self._buffer.writeInteger(0, as: UInt8.self)
```
kmahar, line 61: add a comment that this is null terminating byte?

Sources/BSON/Document.swift
```diff
@@ -28,7 +87,20 @@ public struct Document {
      *
      * - SeeAlso: http://bsonspec.org/
      */
-    public init(fromBSON bson: Data, validate: Bool = true) throws { fatalError("Unimplemented") }
+    public init(fromBSON bson: Data, validate: Bool = true) throws {
```
kmahar, line 90: we'll need to update this based on the rebase. I'm fine with deferring validation to a later PR, but let's create a ticket and add a TODO so we remember to add it.

Sources/BSON/Document.swift
```diff
@@ -73,7 +145,12 @@ public struct Document {
...
      */
     public subscript(key: String) -> BSON? {
-        get { fatalError("Unimplemented") }
+        get {
+            for (docKey, value) in self where docKey == key {
+                return value
+            }
+            return nil
+        }
         set { fatalError("Unimplemented") }
```
kmahar, line 154: I'm also fine with leaving the setter implementation here and below to a future PR and only supporting writing values via the initializer for the moment, but we should make a ticket for adding these

Sources/BSON/DocumentIterator.swift
```diff
@@ -21,14 +21,30 @@ public struct DocumentIterator: IteratorProtocol {
     private var buffer: ByteBuffer
 
     internal init(over buffer: ByteBuffer) {
-        fatalError("Unimplemented")
+        self.buffer = buffer
+        _ = self.buffer.readInteger(as: Int32.self) // put the readerIndex at the first key
     }
 
     /// Advances to the next element and returns it, or nil if no next element exists.
     public mutating func next() -> (String, BSON)? {
```
kmahar, line 29: I think we need to work on error case handling a bit in this method. we can't throw due to protocol restrictions, but in cases where something goes wrong, IMO we should fatal error with a descriptive message rather than just returning nil.

any document we are iterating over here has either been explicitly validated already (via the public initializer) or is a document we ourselves have chosen to bypass validation for.

WDYT @patrickfreed ?

Sources/BSON/BSONValue.swift
```diff
@@ -63,34 +63,44 @@ public enum BSONType: UInt32 {
     case minKey = 0xFF
     /// Special type which compares higher than all other possible BSON element values
     case maxKey = 0x7F
+
+    public var toByte: UInt8 {
```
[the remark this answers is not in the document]
nbbeeken: That would be nice I didn't wanna change it incase there was a reason for it (and like api breakages reasons)

Sources/BSON/Document.swift
nbbeeken: https://jira.mongodb.org/browse/SWIFT-866
adding the todo in the next commit

Sources/BSON/Document.swift
nbbeeken: https://jira.mongodb.org/browse/SWIFT-867
adding the todo in the next commit

Sources/BSON/DocumentIterator.swift
patrickfreed: I think that given the contract you mentioned, `fatalError` here seems okay. It just means that we need to be really careful with the places that we bypass validation (which we should have been doing anyways).

nbbeeken:
Btw the runtime issues made no sense and have since resolved itself 
cannot repro ¯\_(ツ)_/¯

Sources/BSON/Document.swift
```diff
@@ -9,55 +36,115 @@ public struct Document {
...
+        self.keySet = Set(keyValuePairs.map { $0.0 })
+        guard self.keySet.count == keyValuePairs.count else {
+            fatalError("Dictionary \(keyValuePairs) contains duplicate keys")
+        }
+
+        self._buffer = BSON_ALLOCATOR.buffer(capacity: 100)
+
+        guard !self.keySet.isEmpty else {
+            // 5 byte count as Int32 + null byte
+            self._buffer.writeBytes([5, 0, 0, 0] + [0])
```
kmahar, line 61: I think it would be more explicit to do this the way you did in the empty initializer below where you write the int32 5 and then write the null byte. (maybe you can even delegate to that initializer from here?)

Sources/BSON/Document.swift
```diff
@@ -9,55 +36,115 @@ public struct Document {
 
     private var _buffer: ByteBuffer
 
+    internal var size: Int {
+        guard let size = self.buffer.getInteger(at: 0, endianness: .little, as: Int32.self) else {
```
kmahar, line 40: would you ever expect reading the int32 here to fail?

Sources/BSON/Document.swift
```diff
@@ -1,6 +1,33 @@
...
+extension ByteBuffer {
+    @discardableResult
+    internal mutating func writeCString(_ string: String) -> Int {
+        let written = self.writeString(string + "\0")
+        return written
+    }
+
+    internal mutating func readCString() throws -> String? {
+        var string = ""
+        for i in 0..<0xFFFFFE {
```
kmahar, line 16: is this corresponding to max BSON document size? consider introducing a constant for that we can reference throughout the codebase

Sources/BSON/Document.swift
```diff
@@ -1,6 +1,33 @@
...
+                if b[0] == 0 {
+                    return string
+                }
+                guard let character = String(bytes: b, encoding: .utf8) else {
+                    throw InternalError(message: "Cannot decode CString, byte: \(b) at position \(i) as utf8")
+                }
+                string += character
+            }
+        }
+        throw InternalError(message: "Failed to read CString, possibly missing null terminator?")
```
kmahar, line 27: so the only other way this could fail is if your buffer is too long to be valid BSON, right?

nbbeeken: Yep. Although I'll add a condition actually incase the readIndex is in a position where you can't read anything at all.

Sources/BSON/Document.swift
nbbeeken: Not currently we should always have the int32 bytes in the buffer. I do have to unwrap and cast the value nonetheless, is there a cleaner syntax for that here?

Sources/BSON/BSONValue.swift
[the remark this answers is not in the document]
patrickfreed: sgtm

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,184 @@
 import Foundation
 import NIO
 
+/// This shared allocator instance should be used for all underlying `ByteBuffer` creation.
+private let BSON_ALLOCATOR = ByteBufferAllocator()
+/// Maximum BSON document size in bytes
+private let BSON_MAX_SIZE = 0x100000
```
kmahar, line 7: is this actually the max size? I thought the [max size](https://docs.mongodb.com/manual/reference/limits/#BSON-Document-Size) was 16 MB or 16000000 bytes.

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,184 @@
...
+extension ByteBuffer {
+    @discardableResult
+    internal mutating func writeCString(_ string: String) -> Int {
+        let written = self.writeString(string + "\0")
+        return written
+    }
+
+    internal mutating func readCString() throws -> String {
+        let string = try self.getCString(at: self.readerIndex)
+        self.moveReaderIndex(forwardBy: string.utf8.count + 1)
```
kmahar, line 18: do this and `getCString` need to be two separate methods? I only see `getCString` being used here

Sources/BSON/Document.swift
kmahar: id say we should just fatalError as it would be invalid BSON if this fails

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,184 @@
...
      *
      * - Throws:
      *   - `InvalidArgumentError` if the data passed is invalid BSON
      *
      * - SeeAlso: http://bsonspec.org/
      */
-    public init(fromBSON bson: ByteBuffer) throws { fatalError("Unimplemented") }
+    public init(fromBSON bson: ByteBuffer) throws {
+        guard BSONDocument.isValidBSON(bson) else {
+            throw BSONInternalError("Found \(bson) is invalid BSON")
```
kmahar, line 137: i think this should be an invalid argument error

Sources/BSON/BSONErrors.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/// An empty protocol for encapsulating all errors that this package can throw.
+public protocol BSONError: LocalizedError {}
+
+/// An error thrown when the bson library encounters a internal error not caused by the user.
+/// This is usually indicative of a bug in the bson library or system related failure (e.g. memory allocation failure).
+public struct BSONInternalError: BSONError {
```
kmahar, line 8: should update to be namespaced like it is in the driver now?

Sources/BSON/BSONDocument.swift
```diff
@@ -0,0 +1,240 @@
...
+     * Initializes a new `BSONDocument` from the provided BSON data.
+     *
+     * - Throws:
+     *   - `InvalidArgumentError` if the data passed is invalid BSON
+     *
+     * - SeeAlso: http://bsonspec.org/
+     */
+    public init(fromBSON bson: Data) throws {
+        guard BSONDocument.isValidBSON(bson) else {
+            throw BSONError.InvalidArgumentError(message: "Found `\([UInt8](bson))` is invalid BSON")
```
kmahar, line 82: I'm not sure how helpful printing out the byte array is just because it might be very lengthy. 
I'm also thinking it might be better if we just call the method like `validate` and have it throw if the data is not valid. then, depending on why the BSON is invalid, we can provide a more specific error message.

Sources/BSON/BSONInt.swift
```diff
@@ -0,0 +1,35 @@
+import NIO
```
kmahar, line 1: I think for consistency it might be better if we named this file like `Integers+BSONValue.swift`.
and then you'll probably have similar files for like `Date+BSONValue.swift`, etc.
we often name protocol conformance files in that way.

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -0,0 +1,48 @@
+import NIO
+
+extension ByteBuffer {
+    @discardableResult
+    internal mutating func writeCString(_ string: String) -> Int {
+        let written = self.writeString(string + "\0")
+        return written
+    }
+
+    internal mutating func readCString() throws -> String {
```
kmahar, line 10: it's not clear to me why this is split into 3 methods, do we expect to need to use either of `getCString` or `getBSONKey` out of the context of `readCString`?

Sources/BSON/BSONError.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        internal init(value: BSONValue, forKey: String) {
+            self.message =
+                "Failed to set value for key \(forKey) to \(value) with" +
+                " BSON type \(value.bsonType): document too large"
+        }
+    }
+}
+
+internal func wrongIterTypeError(_ iter: BSONDocumentIterator, expected type: BSONValue.Type) -> BSONErrorProtocol {
```
kmahar, line 44: if I'm not mistaken we aren't using this or any of the below helpers yet; can we wait to bring them in until they are being used?

Sources/BSON/BSONDocument.swift
```diff
@@ -0,0 +1,240 @@
+import Foundation
```
kmahar, line 1: can we maybe keep the current name of the file and you can move it after this separately? it's impossible to see the diff in this file because it's all being considered new code

Sources/BSON/ByteBuffer+BSON.swift
nbbeeken: Yea I should've called this out, these helpers are from the work I've done for completing the document iterator, I wanted a get* version of the readCString so that I have an equivalent method that doesn't move the readerIndex
The BSONKey one I can leave out of this PR and include it later but the thinking there was there is code where I need get the the key value including its null byte for equality reasons and so that I don't have to translate the string back to utf8 just to get an accurate length

Sources/BSON/BSONInt.swift
nbbeeken: Yea that makes sense, I did breakout my ByteBuffer extensions into their own file `ByteBuffer+BSON.swift` but since I'm now renaming files to follow this format this would seem to imply it "implements" the BSON enum lol so I changed it to `ByteBuffer+SwiftBSON.swift` since there's no real conformance to list

I did find this tutorial that I can dig up again but basically it suggested when adding methods like this to an outside type you could instead confine your additional methods to a single property

```swift
class BSONHelper { func getCString {...} }
extension ByteBuffer {
  internal var asBSON: BSONHelper
}
``` 
Kinda like this example although there was a trick relating to extending it such that BSONHelper is the same type as ByteBuffer so that `self.` accessors still work. I couldn't get that to work when I was throwing it together to test, does that sound like something we'd like or even matters here? We could collab on the solution if so

nbbeeken: Yea I think its exactly that when the extensions are public, I agree we can keep the current way, and I'll rename 👍

Sources/BSON/ByteBuffer+BSON.swift
kmahar: I think for review purposes it will probably be clearer if we can bring in all these changes incrementally as we use them; otherwise it's hard for reviewers to evaluate the solution in the context it's being used in.
separately, any that are left in should have at least a basic 1-2 sentence docstring added describing their behavior.

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,150 @@
...
 /// A struct representing the BSON document type.
 @dynamicMemberLookup
-public struct Document {
+public struct BSONDocument {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BSON)
 
     private var _buffer: ByteBuffer
 
+    internal var size: Int {
```
kmahar, line 19: I think maybe we could give this a clearer name, perhaps `byteLength`

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+        }
+        // otherwise we're in Xcode, get the bundle's resource path
+        guard let path = Bundle(for: self).resourcePath else {
+            XCTFail("Missing resource path")
+            return ""
+        }
+        return path
+    }
+
+    struct TestError: Error {
```
kmahar, line 23: we have a `TestError` type here and another 10 lines down, are both needed?

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+        let message: String
+    }
+
+    func fail(_ message: String = "This should not happen") -> Never {
```
kmahar, line 27: are we using this anywhere? / if you don't want tests to keep running after failures are encountered the preferred way is to set `continueAfterFailure` to false, like here: https://github.com/mongodb/mongo-swift-driver/blob/26b17296a96ab6a0255bbbf1a86e71565ce3f3c0/Tests/MongoSwiftSyncTests/RetryableReadsTests.swift#L53

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+    if let sd = subdirectory {
+        path += "/\(sd)"
+    }
+    return try FileManager.default
+        .contentsOfDirectory(atPath: path)
+        .filter { $0.hasSuffix(".json") }
+        .map { filename in
+            let url = URL(fileURLWithPath: "\(path)/\(filename)")
+            let data = try Data(contentsOf: url, options: .mappedIfSafe)
+            let jsonResult = try JSONDecoder().decode(T.self, from: data)
```
kmahar, line 59: let's add a todo about switching back to BSONEncoder here, just for the sake of getting more coverage of the encoder

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+        if (33 < byte) && (byte < 126) {
+            string += String(UnicodeScalar(byte))
+        } else {
+            string += "\\x" + String(format: "%02X", byte)
+        }
+    }
+    return string + ", \(String(format: "0x%02X", bytes.count))"
+}
+
+extension BSONDocument: NMBCollection {}
```
kmahar, line 92: why do we conform to this?

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+public extension BSONDocument {
+    func toByteString() -> String {
+        guard let bytes = self.buffer.getBytes(at: 0, length: self.buffer.readableBytes) else {
+            return ""
+        }
+        return BSONTests.toByteString(bytes)
+    }
+
+    func readAllBytes() -> [UInt8] {
+        guard let bytes = self.buffer.getBytes(at: 0, length: self.buffer.readableBytes) else {
```
kmahar, line 73: nit: you could oneline this: `self.buffer.getBytes(at: 0, length: self.buffer.readableBytes ?? []`

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+            return []
+        }
+        return bytes
+    }
+}
+
+public func makeByteString(from bytes: [UInt8]) -> String {
```
kmahar, line 80: this helper doesn't look like it's used anywhere, seems like a duplicate of `toByteString`?

Tests/BSONTests/BSONTests.swift
```diff
@@ -1 +1,143 @@
...
+        if (33 < byte) && (byte < 126) {
+            string += String(UnicodeScalar(byte))
+        } else {
+            string += "\\x" + String(format: "%02X", byte)
+        }
+    }
+    return string
+}
+
+public func toByteString(buffer: ByteBuffer) -> String {
```
kmahar, line 109: I think it would be good if we can make the per-type helpers for this consistent, e.g. maybe all of the per-type methods should be instance methods defined in extensions, like you did for array

Tests/BSONTests/DocumentTests.swift
```diff
@@ -0,0 +1,65 @@
...
+
+final class DocumentTests: BSONTestCase {
+    // This is a test in itself, will fail to compile on unsupported values
+    static let testDoc: BSONDocument = [
+        "int": 0xBAD1DEA,
+        "int32": .int32(32),
+        "int64": .int64(64)
+    ]
+
+    func testInt32Encoding() {
```
kmahar, line 13: do this and the next test give us any coverage we aren't getting now that the int32/int64 corpus tests are running?

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,158 @@
...
      * - Throws:
      *   - `InvalidArgumentError` if the data passed is invalid BSON
      *
      * - SeeAlso: http://bsonspec.org/
      */
-    public init(fromBSON bson: Data) throws { fatalError("Unimplemented") }
+    public init(fromBSON bson: Data) throws {
+        var buffer = BSON_ALLOCATOR.buffer(capacity: bson.count)
+        buffer.writeBytes(bson)
+        buffer.moveReaderIndex(to: 0)
```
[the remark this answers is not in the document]
nbbeeken: While I was debugging an issue it appeared as though the readerIndex was moved to the end, but redebugging shows I missed something so it it already zero here will remove

Tests/BSONTests/BSONTests.swift
nbbeeken: I should document it but is was so I could use Nimble to check count idiomatically
```plain
cannot convert value of type 'BSONDocument' to expected argument type 'NMBCollection?'
expect(DocumentTests.testDoc).to(haveCount(3))
```
I could also just do .count eq 3

Tests/BSONTests/BSONTests.swift
nbbeeken: I cleaned these up and changed the functionality a bit, right now there's no usage cus we didn't need the additional int tests I wrote but they come in handy while developing a BSON type for readable binary. I can remember (or make a ticket) to remove them when all is said and done

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -1,32 +1,59 @@
...
+        self.buffer = buffer
+        // moves readerIndex to first key's type indicator
+        self.buffer.moveReaderIndex(to: 4)
     }
 
     /// Advances to the next element and returns it, or nil if no next element exists.
     public mutating func next() -> (String, BSON)? {
-        fatalError("Unimplemented")
+        guard self.buffer.readableBytes != 0 else {
+            // Iteration has been exhuasted
```
kmahar, line 32: exhuasted -> exhausted

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -1,32 +1,59 @@
...
+            return nil
+        }
+
+        guard let typeByte = self.buffer.readInteger(as: UInt8.self) else {
+            fatalError("BSONDocumentIterator Failed: Cannot read from \(self.buffer)")
+        }
+
+        guard let type = BSONType(rawValue: typeByte), type != .invalid else {
+            if typeByte == 0 {
```
kmahar, line 41: I think it might be a bit clearer to first check if the typeByte is 0 and then return nil, and then handle the invalid case afterward

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// moving the reader index forward by the byte length of string + 1 (for null terminator).
+    internal mutating func readCString() throws -> String {
+        var bytes: [UInt8] = []
+        for _ in 0..<BSON_MAX_SIZE {
+            guard let b = self.readBytes(length: 1) else {
+                throw BSONError.InternalError(message: "Failed to read CString, unable to read byte from \(self)")
+            }
+            guard b[0] != 0 else {
+                guard let string = String(bytes: bytes, encoding: .utf8) else {
+                    throw BSONError.InternalError(message: "Failed to decode BSONKey as UTF8: \(bytes)")
```
kmahar, line 22: `BSONKey` -> `CString`?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -99,124 +98,114 @@ final class BSONCorpusTests: MongoSwiftTestCase {
...
+                    let docFromCB = try BSONDocument(fromBSON: cBData)
+                    expect(docFromCB.toData()).to(equal(cBData))
 
                     // test round tripping through documents
                     // We create an array by reading every element out of the document (and therefore out of the
-                    // bson_t). We then create a new document and append each element of the array to it. Once that
-                    // is done, every element in the original document will have gone from bson_t -> Swift data type
-                    // -> bson_t. At the end, the new bson_t should be identical to the original one. If not, our bson_t
-                    // translation layer is lossy and/or buggy.
-                    let nativeFromDoc = docFromCB.toArray()
```
kmahar, line 149: I think we still want to be using `toArray()` and `init(fromArray:)` here. they are subtly different. 
`toArray()` recursively calls the same thing on sub-document, ensuring we fully decode through all layers of nested data. `Array(...)` will only do that at the top level, and nested documents will not get fully decoded into Swift types.

I think you are missing these at the moment because they are defined in the driver's `DocumentTests.swift`: https://github.com/mongodb/swift-bson/blob/master/Tests/DisabledTests/DocumentTests.swift#L32-L64

they are actually only used by the corpus runner - at some point we moved the corpus runner to its own file but forgot to bring the helpers over. so I think you could just move them over in this PR and then switch back to using them here.

Tests/BSONTests/BSONTests.swift
kmahar: gotcha, didn't know that was a thing! I guess maybe they automatically make `Collection` types conform so we didn't have to do it in the driver

kmahar:
whoops, I accidentally clicked submit too soon but I don't have further comments anyway. few small things left but looking very good!!

Tests/BSONTests/BSONCorpusTests.swift
nbbeeken: Ah good call, I figured this was API on the actual BSONDocument I hadn't done yet, I pulled in the helpers but I can't use them until I make document's mutable which has a ticket so marked this TODO

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,156 @@
...
      */
-    public init(fromBSON bson: Data) throws { fatalError("Unimplemented") }
+    public init(fromBSON bson: Data) throws {
+        var buffer = BSON_ALLOCATOR.buffer(capacity: bson.count)
+        buffer.writeBytes(bson)
+        try BSONDocument.validate(buffer)
+        self = BSONDocument(fromUnsafeBSON: buffer)
+    }
+
+    internal init(fromUnsafeBSON bson: Data) {
```
patrickfreed, line 88: is this initializer necessary?

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,156 @@
...
      *   - `InvalidArgumentError` if the data passed is invalid BSON
      *
      * - SeeAlso: http://bsonspec.org/
      */
-    public init(fromBSON bson: Data) throws { fatalError("Unimplemented") }
+    public init(fromBSON bson: Data) throws {
+        var buffer = BSON_ALLOCATOR.buffer(capacity: bson.count)
+        buffer.writeBytes(bson)
+        try BSONDocument.validate(buffer)
+        self = BSONDocument(fromUnsafeBSON: buffer)
```
patrickfreed, line 85: these two lines can be replaced with a call to the `fromBSON: ByteBuffer` initializer.

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,156 @@
...
+        self._buffer = bson
+        for (key, _) in self {
+            self.keySet.insert(key)
+        }
+    }
 
-    /// The keys in this `Document`.
-    public var keys: [String] { fatalError("Unimplemented") }
+    /// The keys in this `BSONDocument`.
+    public var keys: [String] { self.map { key, _ in key } }
```
patrickfreed, line 117: this `map` decodes the entire document just to get the keys, which we already store in a separate data structure. That structure is undordered though, so I don't think we can use it here. We could possibly replace the `Set` of keys with an `NSMutableOrderedSet` as a performance optimization in the future. That said, perf is out of scope for this work, so I think it's fine as-is.

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,156 @@
...
 
     /// The number of (key, value) pairs stored at the top level of this document.
-    public var count: Int { fatalError("Unimplemented") }
+    public var count: Int { self.keySet.count }
 
     /// A copy of the `ByteBuffer` backing this document, containing raw BSON data. As `ByteBuffer`s implement
     /// copy-on-write, this copy will share byte storage with this document until either the document or the returned
     /// buffer is mutated.
-    public var buffer: ByteBuffer { fatalError("Unimplemented") }
+    public var buffer: ByteBuffer { ByteBuffer(self._buffer.readableBytesView) }
```
patrickfreed, line 128: Given that this property is making an explicit copy, I don't think the CoW behavior is actually happening. To enable CoW, just return the buffer itself (e.g. `self._buffer`).

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -0,0 +1,35 @@
+import NIO
+
+extension Int32: BSONValue {
```
patrickfreed, line 3: this could all use explicit acl's

Sources/BSON/Document.swift
```diff
@@ -94,58 +173,63 @@ public struct Document {
...
-    public static func == (lhs: Document, rhs: Document) -> Bool {
+extension BSONDocument: Equatable {
+    public static func == (lhs: BSONDocument, rhs: BSONDocument) -> Bool {
         fatalError("Unimplemented")
     }
 }
 
-extension Document: BSONValue {
-    var bsonType: BSONType { fatalError("Unimplemented") }
+extension BSONDocument: BSONValue {
```
patrickfreed, line 223: ditto explicit acl's

Sources/BSON/Document.swift
```diff
@@ -1,77 +1,156 @@
...
     internal init(_ elements: [BSON]) { fatalError("Unimplemented") }
 
-    internal init(keyValuePairs: [(String, BSON)]) { fatalError("Unimplemented") }
+    internal init(keyValuePairs: [(String, BSON)]) {
+        self.keySet = Set(keyValuePairs.map { $0.0 })
+        guard self.keySet.count == keyValuePairs.count else {
+            fatalError("Dictionary \(keyValuePairs) contains duplicate keys")
+        }
+
+        self._buffer = BSON_ALLOCATOR.buffer(capacity: 100)
```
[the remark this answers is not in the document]
nbbeeken: Yes the capacity grows on write, I just chose 100 at random pushing optimization work here for a later day. I think I should actually make this zero and let the first write call make whatever default allocation swiftNIO thinks is best

Sources/BSON/Document.swift
nbbeeken: I changed it but just as an experiment I verified there's no change, I guess the initializer is smart enough not to touch the underlying storage unless necessary.

PR #6, opened 2020-05-19
SWIFT-840: Decimal 128 Support

nbbeeken:
(Finally _amiright_)

This adds support to create Decimal128 values from a string and round trip the values to and from BSON.
Note Decimal128.toString is missing right now, this is something that should probably be a part of extJSON work.

### DIFF
--- a/Sources/BSON/BSON.swift
+++ b/Sources/BSON/BSON.swift
@@ -12,6 +12,9 @@ public enum BSON {
     /// A BSON int64.
     case int64(Int64)
 
+    /// A BSON Decimal128
+    case decimal128(BSONDecimal128)
+
     /// A BSON Array
     indirect case array([BSON])
 
@@ -93,6 +96,14 @@ extension BSON {
     /// If this `BSON` is an `.int64`, return it as an `Int64`. Otherwise, return nil.
     public var int64Value: Int64? {
         guard case let .int64(i) = self else {
+            return nil
+        }
+        return i
+    }
+
+    /// If this `BSON` is an `.decimal128`, return it as an `BSONDecimal128`. Otherwise, return nil.
+    public var decimal128Value: BSONDecimal128? {
+        guard case let .decimal128(i) = self else {
             return nil
         }
         return i
@@ -266,6 +277,7 @@ extension BSON {
         .document: BSONDocument.self,
         .int32: Int32.self,
         .int64: Int64.self,
+        .decimal128: BSONDecimal128.self,
         .bool: Bool.self,
         .string: String.self,
         .double: Double.self,
@@ -293,6 +305,8 @@ extension BSON {
         case let .int32(v):
             return v
         case let .int64(v):
+            return v
+        case let .decimal128(v):
             return v
         case let .array(v):
             return v
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -308,7 +308,13 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
         // the product is the significandHiDigits "shifted" up by 17 decimal places
         // we can then add the significandLoDigits to the product to ensure that we have a correctly formed significand
         var significand = UInt128.multiply(significandHiDigits, by: Self.decimalShift17Zeroes)
-        significand.lo += significandLoDigits
+
+        let (result, didOverflow) = significand.lo.addingReportingOverflow(significandLoDigits)
+        significand.lo = result
+
+        if didOverflow {
+            significand.hi += 1
+        }
 
         let biasedExponent = UInt64(exponent + Self.exponentBias).getLeastSignificantBits(Self.exponentLength)
 
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -339,7 +339,7 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
         let combination = (self.value.hi >> 58) & 0x1F
         if (combination >> 3) == 0b11 {
             if combination == COMBINATION_INFINITY {
-                return self.isNegative ? "-" : "" + "Infinity"
+                return (self.isNegative ? "-" : "") + "Infinity"
             }
             if combination == COMBINATION_NAN {
                 return "NaN"
@@ -385,9 +385,6 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
                 var (quotient, remainder) = UInt128.divideBy1Billion(significand128)
                 significand128 = quotient
                 /* We now have the 9 least significant digits. */
-                if remainder == 0 {
-                    continue
-                }
                 for _ in 0...8 {
                     significand.insert(remainder % 10, at: 0)
                     remainder /= 10
@@ -395,11 +392,15 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
             }
         }
 
-        /* Scientific - [-]d.dddE(+/-)dd or [-]dE(+/-)dd */
+        if !isZero, let firstNonZero = significand.firstIndex(where: { $0 != 0 }) {
+            significand = [Int](significand.suffix(from: firstNonZero))
+        }
+
+        /* Scientific - [-]d.ddde(+/-)dd or [-]de(+/-)dd */
         /* Regular    - ddd.ddd */
 
         /*
-         * The scientific exponent checks are dictated by the string conversion
+         * The adjusted_exponent checks are dictated by the string conversion
          * specification.
          *
          * We must check exponent > 0, because if this is the case, the number
@@ -415,8 +416,8 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
             representation += String(significand[0], radix: 10)
             representation += significand.count > 1 ? "." : ""
             representation += significand[1..<significand.count].map { String($0, radix: 10) }.joined(separator: "")
-            representation += "e"
-            representation += String(format: "%+d", exponent)
+            representation += "E"
+            representation += String(format: "%+d", adjusted_exponent)
         } else {
             // Regular format
             guard exponent != 0 else {
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -24,7 +24,7 @@ private extension UInt64 {
     }
 
     func getLeastSignificantBits<T: FixedWidthInteger>(_ length: T) -> Self {
-        self.getBits((63 - Int(length))...63)
+        self.getBits((63 - Int(length - 1))...63)
     }
 
     /// Gets this number's bits shifting the value down to the LSB
@@ -121,12 +121,12 @@ internal struct UInt128: Equatable, Hashable {
 public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
     // swiftlint:disable line_length
     private static let digitsRegex = #"(?:\d+)"#
-    private static let indicatorRegex = #"(?:e|E)"#
+    private static let indicatorRegex = #"(?:e)"#
     private static let signRegex = #"[+-]"#
     private static let infinityRegex = #"infinity|inf"#
     private static let decimalRegex = "\(digitsRegex)\\.\(digitsRegex)?|\\.?\(digitsRegex)"
     private static let nanRegex = #"NaN"#
-    private static let exponentRegex = "\(indicatorRegex)(\(signRegex))(\(digitsRegex))"
+    private static let exponentRegex = "\(indicatorRegex)(\(signRegex))?(\(digitsRegex))"
     private static let numericValueRegex = "(\(signRegex))?(?:(\(decimalRegex))(?:\(exponentRegex))?|(\(infinityRegex)))"
     public static let decimal128Regex = "\(numericValueRegex)|(\(nanRegex))"
     // swiftlint:enable line_length
@@ -382,7 +382,7 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
         // significand prefix (implied bits) combined with removing the combination and sign fields
         significand128.hi = UInt64(
             significandPrefix.getLeastSignificantBits(4) << (Self.trailingSignificandLength - 64))
-            | self.value.hi.getLeastSignificantBits(45)
+            | self.value.hi.getLeastSignificantBits(Self.trailingSignificandLength - 64)
         significand128.lo = self.value.lo
 
         // make a base 10 digits array from significand
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -320,8 +320,13 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
 
         var value = UInt128()
 
+        // Normally Decimal(K) encodings would conditionally modify the combination field here
+        // based on the most significant bits of the significand.
+        // Decimal128 doesn't actually need the extra implicit bits.
+        // The entirety of decimal128's range can fit by just encoding the exponent and significand as-is.
+
         value.hi |= biasedExponent << (63 - Self.exponentLength)
-        value.hi |= significand.hi.getLeastSignificantBits(60)
+        value.hi |= significand.hi
 
         value.lo = significand.lo
 
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -9,7 +9,7 @@ private extension UInt64 {
     /// the bits are indexed from MSB at 0 to LSB at 63
     /// example: (0b11010).getBitsUnshifted(0..<4) == 0b110_00
     /// (note: the bits before the _ are the one's gotten)
-    func getBitsUnshifted(_ range: ClosedRange<Int>) -> Self {
+    func getBitsUnshifted<T: FixedWidthInteger>(_ range: ClosedRange<T>) -> Self {
         guard range.lowerBound >= 0 else {
             fatalError("BSONDecimal128: Your range should be bound between [0, 63] was \(range)")
         }
@@ -30,7 +30,7 @@ private extension UInt64 {
     /// Gets this number's bits shifting the value down to the LSB
     /// the bits are indexed from MSB at 0 to LSB at 63
     /// example: (0b11010).getBitsUnshifted(0..<4) == 0b110
-    func getBits(_ range: ClosedRange<Int>) -> Self {
+    func getBits<T: FixedWidthInteger>(_ range: ClosedRange<T>) -> Self {
         var value = self.getBitsUnshifted(range)
         value >>= (63 - range.upperBound)
         return value
@@ -131,18 +131,21 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
     public static let decimal128Regex = "\(numericValueRegex)|(\(nanRegex))"
     // swiftlint:enable line_length
 
-    // The precision of the Decimal128 format
+    /// The precision of the Decimal128 format
     private static let maxSignificandDigits = 34
-    // NOTE: the min and max values are adjusted for when the decimal point is rounded out
-    // e.g, 1.000...*10^-6143 == 1000...*10^-6176
-    // In the spec exp_max is 6144 so we use 6111
+    /// NOTE: the min and max values are adjusted for when the decimal point is rounded out
+    /// e.g, 1.000...*10^-6143 == 1000...*10^-6176
+    /// In the spec exp_max is 6144 so we use 6111
     private static let exponentMax = 6111
-    // In the spec exp_min is -6134 so we use -6176
+    /// In the spec exp_min is -6134 so we use -6176
     private static let exponentMin = -6176
-    // The sum of the exponent and a constant (bias) chosen to make the biased exponent’s range non-negative.
+    /// The sum of the exponent and a constant (bias) chosen to make the biased exponent’s range non-negative.
     private static let exponentBias = 6176
 
+    /// Length in bits of the exponent field
     private static let exponentLength: UInt64 = 14
+    /// Length in bits of the combination field
+    private static let combinationLength: UInt64 = 3
 
     private static let decimalShift17Zeroes: UInt64 = 100_000_000_000_000_000
 
@@ -371,12 +374,16 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
             }
             // The number is neither NaN nor Inf
             // Decimal interchange floating-point formats c,2,ii
-            exponent = Int(self.value.hi.getBits(3...16))
+            exponent = Int(self.value.hi.getBits(3...(Self.exponentLength + 2)))
             significandPrefix = Int(self.value.hi.getBit(20) + 0b1000)
         } else {
             // Decimal interchange floating-point formats c,2,i
-            exponent = Int(self.value.hi.getBits(1...14))
-            significandPrefix = Int(self.value.hi.getBits(15...17))
+            exponent = Int(self.value.hi.getBits(1...Self.exponentLength))
+            significandPrefix = Int(
+                self.value.hi.getBits(
+                    (Self.exponentLength + 1)...(Self.exponentLength + Self.combinationLength)
+                )
+            )
         }
 
         exponent -= Self.exponentBias
@@ -388,7 +395,7 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
         significand128.lo = self.value.lo
 
         // make a base 10 digits array from significand
-        var significand = [Int]()
+        var significandDigits = [Character]()
 
         var isZero = false
 
@@ -406,25 +413,25 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
         }
 
         if isZero {
-            significand = [0]
+            significandDigits = ["0"]
         } else {
             for _ in 0...3 {
                 var (quotient, remainder) = significand128.divideBy1Billion()
                 significand128 = quotient
                 // We now have the 9 least significant digits.
                 for _ in 0...8 {
-                    significand.insert(remainder % 10, at: 0)
+                    significandDigits.insert(Character(String(remainder % 10, radix: 10)), at: 0)
                     remainder /= 10
                 }
             }
         }
 
-        if !isZero, let firstNonZero = significand.firstIndex(where: { $0 != 0 }) {
-            significand = [Int](significand.suffix(from: firstNonZero))
+        if !isZero, let firstNonZero = significandDigits.firstIndex(where: { $0 != "0" }) {
+            significandDigits = [Character](significandDigits.suffix(from: firstNonZero))
         }
 
-        // Scientific - [-]d.ddde(+/-)dd or [-]de(+/-)dd
-        // Regular    - ddd.ddd
+        // Exponential - [-]d.ddde(+/-)dd or [-]de(+/-)dd
+        // Regular     - ddd.ddd
 
         /*
          * The adjusted_exponent checks are dictated by the string conversion
@@ -437,41 +444,39 @@ public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
          */
         var representation = self.isNegative ? "-" : ""
 
-        let adjusted_exponent = exponent + (significand.count - 1)
+        let adjusted_exponent = exponent + (significandDigits.count - 1)
         if exponent > 0 || adjusted_exponent < -6 {
-            // Scientific format
-            representation += String(significand[0], radix: 10)
-            representation += significand.count > 1 ? "." : ""
-            representation += significand[1..<significand.count].map { String($0, radix: 10) }.joined(separator: "")
+            // Exponential format
+            representation += String(significandDigits[0])
+            representation += significandDigits.count > 1 ? "." : ""
+            representation += String(significandDigits[1..<significandDigits.count])
             representation += "E"
             representation += String(format: "%+d", adjusted_exponent)
         } else {
             // Regular format
             guard exponent != 0 else {
-                representation += significand.map { String($0, radix: 10) }.joined(separator: "")
+                representation += String(significandDigits)
                 return representation
             }
 
-            var pointPosition = significand.count + exponent
+            var pointPosition = significandDigits.count + exponent
 
             if pointPosition > 0 {
                 // number isn't a fraction
-                for _ in 0..<pointPosition {
-                    representation += String(significand[0], radix: 10)
-                    significand = Array(significand.dropFirst())
-                }
+                representation += String(significandDigits[0..<pointPosition])
+                significandDigits = Array(significandDigits[pointPosition...])
             } else {
                 representation += "0"
             }
 
             representation += "."
 
-            while pointPosition < 0 {
-                representation += "0"
-                pointPosition += 1
+            if pointPosition < 0 {
+                representation += String(repeating: "0", count: abs(pointPosition))
+                pointPosition += abs(pointPosition)
             }
 
-            representation += significand.map { String($0, radix: 10) }.joined(separator: "")
+            representation += String(significandDigits)
         }
         return representation
     }
--- a/Sources/BSON/BSONDecimal128.swift
+++ b/Sources/BSON/BSONDecimal128.swift
@@ -35,22 +35,22 @@ internal struct UInt128: Equatable, Hashable {
         let rightLo = right & 0xFFFF_FFFF
 
         var productHi = leftHi * rightHi
-        var productMid0 = leftHi * rightLo
+        let productMid0 = leftHi * rightLo
         let productMid1 = leftLo * rightHi
         var productLo = leftLo * rightLo
 
         productHi += productMid0 >> 32
-        productMid0 = (productMid0 & 0xFFFF_FFFF) + productMid1 + (productLo >> 32)
+        let productPartial = (productMid0 & 0xFFFF_FFFF) + productMid1 + (productLo >> 32)
 
-        productHi += (productMid0 >> 32)
-        productLo = (productMid0 << 32) + (productLo & 0xFFFF_FFFF)
+        productHi += (productPartial >> 32)
+        productLo = (productPartial << 32) + (productLo & 0xFFFF_FFFF)
 
         return UInt128(hi: productHi, lo: productLo)
     }
 
     internal static func divideBy1Billion(_ numerator: UInt128) -> (quotient: UInt128, remainder: Int) {
         // swiftlint:disable:previous cyclomatic_complexity
-        let DIVISOR: UInt64 = 1000 * 1000 * 1000
+        let denominator: UInt64 = 1000 * 1000 * 1000
         var remainder: UInt64 = 0
         var quotient = numerator
 
@@ -73,14 +73,14 @@ internal struct UInt128: Equatable, Hashable {
             remainder += quotient_i
             // quotient[i] = Int(remainder / DIVISOR)
             switch i {
-            case 0: quotient.hi = (((remainder / DIVISOR) << 32) | quotient.hi & 0x0000_0000_FFFF_FFFF)
-            case 1: quotient.hi = (((remainder / DIVISOR) & 0xFFFF_FFFF) | quotient.hi & 0xFFFF_FFFF_0000_0000)
-            case 2: quotient.lo = (((remainder / DIVISOR) << 32) | quotient.lo & 0x0000_0000_FFFF_FFFF)
-            case 3: quotient.lo = (((remainder / DIVISOR) & 0xFFFF_FFFF) | quotient.lo & 0xFFFF_FFFF_0000_0000)
+            case 0: quotient.hi = (((remainder / denominator) << 32) | quotient.hi & 0x0000_0000_FFFF_FFFF)
+            case 1: quotient.hi = (((remainder / denominator) & 0xFFFF_FFFF) | quotient.hi & 0xFFFF_FFFF_0000_0000)
+            case 2: quotient.lo = (((remainder / denominator) << 32) | quotient.lo & 0x0000_0000_FFFF_FFFF)
+            case 3: quotient.lo = (((remainder / denominator) & 0xFFFF_FFFF) | quotient.lo & 0xFFFF_FFFF_0000_0000)
             default: _ = ()
             }
             /* Store the remainder */
-            remainder %= DIVISOR
+            remainder %= denominator
         }
 
         return (quotient: quotient, remainder: Int(remainder & 0xFFFF_FFFF))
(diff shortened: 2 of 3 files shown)


Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -3,6 +3,7 @@ import Foundation
 import Nimble
 import XCTest
 
+@available(OSX 10.13, *)
```
kmahar, line 6: why did we need to add this?

Sources/BSON/BSON.swift
```diff
@@ -53,6 +56,14 @@ public extension BSON {
         }
         return d
     }
+
+    /// If this `BSON` is a `.document`, return it as a `BSONDocument`. Otherwise, return nil.
```
kmahar, line 60: docstring needs to be updated

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,283 @@
+import Foundation
+import NIO
+
+public struct UInt128: Equatable, Hashable {
```
kmahar, line 4: I think this type should probably be internal. the Decimal128 type can just expose the standardized string representation.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,283 @@
...
+                }
+                digits.append(digit - asciiZero)
+            }
+        } else {
+            digits.append(digitsFromRepr[0] - asciiZero)
+        }
+        return digits
+    }
+
+    public static func read(from buffer: inout ByteBuffer) throws -> BSON {
```
kmahar, line 268: since the main type definition is already pretty long, consider moving the BSONValue conformance into a extension in the same file just to make things a bit more organized, like

```swift
extension BSONDecimal128: BSONValue {
   // methods here
}
```

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,283 @@
...
+    /// Determines if the value is Not a Number
+    public var isNaN: Bool { BSONDecimal128.NAN == self.value }
+    /// Determines if the value is Infinity
+    public var isInf: Bool { BSONDecimal128.INFINITY == self.value || BSONDecimal128.NEG_INFINITY == self.value }
+
+    internal init(fromUInt128 value: UInt128) {
+        self.value = value
+    }
+
+    public init(fromString repr: String) throws {
```
kmahar, line 94: this needs to match the initializer that is in the driver now, which is `public init(_ data: String) throws`

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    public static var bsonType: BSONType { .decimal128 }
+
+    public var bson: BSON { .decimal128(self) }
+
+    public var description: String { fatalError("Unimplemented") }
+
+    /// Holder for raw decimal128 value
+    public var value: UInt128
```
kmahar, line 83: let's stick with the public API that was agreed upon in the design doc (so don't expose the next 3 properties). I appreciate the effort to add more potentially useful API, but making the native type usable for computation, comparisons etc. and exposing a UInt128 type is is a whole can of worms that I think is best deferred to a potential separate project for a number of reasons:
1. a `UInt128` type could easily be added to the standard library down the road, in which case we'd want to switch our type to wrap that. publicly exposing this implementation detail boxes us into keeping and maintaining our custom type until a major version release 
2. Arguably interop with `Decimal` (despite being lossy) might be more appealing to users than a custom type for doing computations on
3. This type is not used very widely anyway, so I don't know that maintaining such an API is worth the investment

Tests/BSONTests/BSONCorpusTests.swift
nbbeeken: It was there because I originally wrote my regexes using named capture groups which need that annotation, I went to just using the index since it got frustrating that the whole call chain such an annotation, and that's something we probably wouldn't want to pass along to users. 
Anyway we don't need it anymore, good catch!

Sources/BSON/BSON.swift
```diff
@@ -53,6 +56,14 @@ extension BSON {
         }
         return d
     }
+
+    /// If this `BSON` is a `.document`, return it as a `BSONDecimal128`. Otherwise, return nil.
```
kmahar, line 60: `.document` -> `.decimal128`

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+         * BSON_DECIMAL128_STRING 42
+         */
+        guard !self.isNaN else {
+            return "NaN"
+        }
+        /// wip wip, eta june 10th morning
+    }
+}
+
+extension BSONDecimal128: BSONValue {
```
kmahar, line 287: all the methods/properties in here should be `internal`, since they are part of the internal `BSONValue` protocol and not something users should be using directly

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+    private enum REGroups: Int, CaseIterable {
+        case sign = 1
+        case decimalPart = 2
+        case exponentSign = 3
+        case exponentPart = 4
+        case infinity = 5
+        case nan = 6
+    }
+
+    public enum BSONDecimal128Errors: Error {
```
kmahar, line 68: I think switching the errors thrown to a new type is a breaking change for the driver, so we probably can't use this approach. I'd suggest defining some constants corresponding to each of these cases that are `InvalidArgumentError`s (or whatever the appropriate type would be given the context you're throwing them in)

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+
+        var sign = 1
+        let signRange: NSRange = match.range(at: REGroups.sign.rawValue)
+        if signRange.location != NSNotFound, let range = Range(signRange, in: data) {
+            sign = String(data[range]) == "-" ? -1 : 1
+        }
+
+        let isNaN = match.range(at: REGroups.nan.rawValue)
+        if isNaN.location != NSNotFound {
+            self.value = UInt128(hi: 0x7C00_0000_0000_0000, lo: 0)
```
kmahar, line 110: can we use your static NaN constant here instead?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+            return
+        }
+
+        let isInfinity = match.range(at: REGroups.infinity.rawValue)
+        if isInfinity.location != NSNotFound {
+            if sign < 0 {
+                self.value = UInt128(hi: 0xF800_0000_0000_0000, lo: 0)
```
kmahar, line 117: same q about using constants here and a few lines down

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+    /// Determines if the value is Not a Number
+    public var isNaN: Bool { ((self.value.hi & 0x7F00_0000_0000_0000) >> 56) == 0b11111 }
+    /// Determines if the value is Infinity
+    public var isInf: Bool { ((self.value.hi & 0x7F00_0000_0000_0000) >> 56) == 0b11110 }
+
+    internal init(fromUInt128 value: UInt128) {
+        self.value = value
+    }
+
+    public init(_ data: String) throws {
```
kmahar, line 94: are we testing this logic at all right now? I don't think the corpus tests actually exercise this initializer

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,307 @@
...
+            throw BSONDecimal128Errors.Rounding
+        }
+
+        guard digits.count <= BSONDecimal128.PRECISION_DIGITS else {
+            throw BSONDecimal128Errors.Overflow
+        }
+
+        var significand = UInt128(hi: 0x0, lo: 0x0)
+
+        if digits.isEmpty {
```
kmahar, line 192: aren't they already initialized to 0?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+    public static let DECIMAL128_RE = "\(NUMERIC_VALUE_RE)|(\(NAN_RE))"
+
+    private static let PRECISION_DIGITS = 34
+    // NOTE: the min and max values are adjusted for when the decimal point is rounded out
+    // e.g, 0.000...*10^-6143 == 0.000...*10^-6176
+    // In the spec exp_max is 6144 so we use 6111
+    private static let EXPONENT_MAX = 6111
+    // In the spec exp_min is -6134 so we use -6176
+    private static let EXPONENT_MIN = -6176
+    private static let EXPONENT_BIAS = 6176
```
patrickfreed, line 110: what is the bias / needs a comment?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+                continue
+            }
+
+            if digits.last != 0 {
+                // We don't end in a zero and our exponent is too small
+                throw BSONError.InvalidArgumentError(message: "Underflow Error: Value too small")
+            }
+        }
+
+        guard (exponent >= BSONDecimal128.EXPONENT_MIN) && (exponent <= BSONDecimal128.EXPONENT_MAX) else {
```
patrickfreed, line 223: it might be clearer to say the following (up to you though):
```swift
guard BSONDecmial128.EXPONENT_MIN...BSONDecimal128.EXPONENT_MAX.contains(exponent) else { ... }
```

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+        if digits.isEmpty {
+            significand.hi = 0
+            significand.lo = 0
+        }
+
+        let lo_digits = Array(digits.suffix(17))
+        let hi_digits = Array(digits.dropLast(17))
+
+        if !lo_digits.isEmpty {
+            significand.lo = UInt64(lo_digits[0])
```
patrickfreed, line 244: may be useful to abstract this to a function so it can be re-used below.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+        if product.lo < significand.lo {
+            product.hi += 1
+        }
+
+        let biased_exponent = exponent + BSONDecimal128.EXPONENT_BIAS
+
+        self.value = UInt128()
+
+        // Encode combination, exponent, and significand.
+        if (product.hi >> 49) & 1 == 1 {
```
patrickfreed, line 271: can the inline comments be a bit more explicit about what's happening here?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+        if sign < 0 {
+            self.value.hi |= 0x8000_0000_0000_0000
+        }
+    }
+
+    private static func convertToDigitsArray(_ decimalString: String) throws -> [UInt8] {
+        var leadingZero = true
+        var digits: [UInt8] = []
+
+        let asciiZero = Character("0").asciiValue ?? 48
```
patrickfreed, line 292: these can be `static let` that use the `!` instead of nil coercion.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+        let asciiNine = Character("9").asciiValue ?? 57
+        let asciiPoint = Character(".").asciiValue ?? 46
+
+        let digitsFromRepr = [UInt8](decimalString.utf8)
+        if digitsFromRepr.count > 1 {
+            for digit in digitsFromRepr {
+                if digit == asciiPoint {
+                    continue
+                }
+                if (digit < asciiZero) && (digit > asciiNine) {
```
patrickfreed, line 302: ditto using a range here if you think it improves readability.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,473 @@
...
+         * 1  for the sign
+         * 35 for digits and radix
+         * 2  for exponent indicator and sign
+         * 4  for exponent digits
+         * BSON_DECIMAL128_STRING 42
+         */
+        var exponent: Int
+        var sig_prefix: Int
+
+        let COMBINATION_INFINITY = 30
```
patrickfreed, line 336: some explanation of these constants would be nice, also specifying them as `static let`

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,476 @@
...
+        var digits = try Self.convertToDigitsArray(decimalPart)
+
+        var exponent = 0
+        let exponentPartRange = match.range(at: REGroups.exponentPart.rawValue)
+        if exponentPartRange.location != NSNotFound, let range = Range(exponentPartRange, in: data) {
+            exponent = exponentSign * (Int(data[range]) ?? 0)
+        }
+        if let pointIndex = decimalPart.firstIndex(of: ".") {
+            exponent -= decimalPart.distance(from: pointIndex, to: decimalPart.endIndex) - 1
+            if exponent < Self.exponentMin {
```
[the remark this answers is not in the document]
nbbeeken: They're separate concerns I added a comment hopefully making it a bit clearer, but this subtraction from the exponent has already implicitly happened when we turned the mantissa into an array of digits, this line is that calculation being done explicitly. The while loop below handles rounding.

example: 
- we are given `"1.2340e10"` exponent is == 10
- `convertToDigitsArray` will return `[1,2,3,4,0]` and this subtraction will make exponent == 6
  - It's now as if we are encoding the number `"12340e6"`
  - trailing zeroes are preserved so precision is kept

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+    private static let infinityRegex = #"Infinity|Inf|infinity|inf"#
+    private static let decimalRegex = "\(digitsRegex)\\.\(digitsRegex)?|\\.?\(digitsRegex)"
+    private static let nanRegex = #"NaN"#
+    private static let exponentRegex = "\(indicatorRegex)(\(signRegex))(\(digitsRegex))"
+    private static let numericValueRegex = "(\(signRegex))?(?:(\(decimalRegex))(?:\(exponentRegex))?|(\(infinityRegex)))"
+    public static let decimal128Regex = "\(numericValueRegex)|(\(nanRegex))"
+    // swiftlint:enable line_length
+
+    // The precision of the Decimal128 format
+    private static let significandDigits = 34
```
kmahar, line 104: perhaps this should be `maxSignificandDigits`?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        var significandHiDigits = UInt64()
+        var significandLoDigits = UInt64()
+
+        let loDigits = Array(digits.suffix(Self.significandDigits / 2))
+        let hiDigits = Array(digits.dropLast(Self.significandDigits / 2))
+
+        if !loDigits.isEmpty {
+            significandLoDigits = UInt64(loDigits[0])
```
kmahar, line 251: might be worth extracting the logic that is used both here and for highDigits, e.g. into an `[UInt8]` extension method like `toUInt64()` or something

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+    }
+}
+
+public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
+    // swiftlint:disable line_length
+    private static let digitsRegex = #"(?:\d+)"#
+    private static let indicatorRegex = #"(?:e|E)"#
+    private static let signRegex = #"[+-]"#
+    private static let infinityRegex = #"Infinity|Inf|infinity|inf"#
+    private static let decimalRegex = "\(digitsRegex)\\.\(digitsRegex)?|\\.?\(digitsRegex)"
```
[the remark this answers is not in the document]
nbbeeken: Yea you have to escape the dot since we want to match on a `.` literal and then you have to escape the escape

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+    private static let nanRegex = #"NaN"#
+    private static let exponentRegex = "\(indicatorRegex)(\(signRegex))(\(digitsRegex))"
```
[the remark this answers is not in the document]
nbbeeken: Regexes are hard to make readable especially with all the escaping, I tried to chunk it up to look like the grammar given in the spec
`"(\(signRegex))?(?:(\(decimalRegex))(?:\(exponentRegex))?|(\(infinityRegex)))"`
Anyway this regex below conditionally includes the exponentRegex the question mark before the bar `|`

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+
+    /// Determines if the value is 0
+    private var isNegative: Bool { (self.value.hi >> 63) == 1 }
+
+    /// Indicators in the combination field that determine number type
+    private static let combinationNaN = 0b11111
+    private static let combinationInfinity = 0b11110
+
+    /// Determines if the value is Not a Number
+    private var isNaN: Bool { ((self.value.hi >> 58) & 0x1F) == Self.combinationNaN }
```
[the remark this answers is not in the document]
nbbeeken: Updated the comments, I switched to this method because ieee spec says the rest of NaN can be filled with junk, its all don't care values, and infinity I think is more strict but it felt right to match them up

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
nbbeeken: That's outside the precision of Decimal128 I believe and it would clamp / round to a minimum

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+            )
+        }
+
+        let significandLoDigits = [UInt8](digits.suffix(Self.maxSignificandDigits / 2)).decimalDigitsToUInt64()
+        let significandHiDigits = [UInt8](digits.dropLast(Self.maxSignificandDigits / 2)).decimalDigitsToUInt64()
+
+        // Multiply by one hundred quadrillion (note the seventeen zeroes)
+        // the product is the significandHiDigits "shifted" up by 17 decimal places
+        // we can then add the significandLoDigits to the product to ensure that we have a correctly formed significand
+        var product = UInt128.multiply(significandHiDigits, by: Self.decimalShift17Zeroes)
```
patrickfreed, line 297: can we rename this variable to `significand`? `product` is pretty generic.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+            product.hi += 1
+        }
+
+        let biasedExponent = exponent + Self.exponentBias
+
+        self.value = UInt128()
+
+        // The most significant bit of the significand determines the format
+        // Encode combination, exponent, and significand.
+        if product.hi.getBit(14) == 1 {
```
patrickfreed, line 311: how is this 14 determined to be the MSB of the significand?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+            // The significand has the implicit (0b100) at the
+            // beginning of the trailing significand field
+
+            // Ensure we encode '0b11' into bits 1 to 3
+            self.value.hi |= (0b11 << 61)
+            self.value.hi |= UInt64(biasedExponent & Self.exponentMask) << 47
+            self.value.hi |= product.hi.getBits(5...63)
```
patrickfreed, line 318: given that 14 is the msb, wouldnt everything < 14 be 0?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+        } else {
+            // The significand has the implicit (0b0) at the
+            // beginning of the trailing significand field
+            self.value.hi |= UInt64(biasedExponent & Self.exponentMask) << 49
```
patrickfreed, line 322: now that we have these nice methods for doing masks, I think the constant mask suggestion may no longer be the best approach.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+            self.value.hi |= product.hi.getBits(3...63)
+        }
+
+        self.value.lo = product.lo
+
+        if sign < 0 {
+            self.value.hi |= 0x8000_0000_0000_0000
```
patrickfreed, line 329: this could be replaced by a similar `setBit` method.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
+import Foundation
+import NIO
+
+private extension UInt64 {
+    var upper32bits: Self { self.getBits(0...31) }
+    var lower32bits: Self { self.getBits(32...63) }
+
+    /// Gets this number's bits without shifting the value down to the LSB
+    /// the bits are indexed from MSB at 0 to LSB at 63
```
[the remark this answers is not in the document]
patrickfreed: The spec is actually written in MSB 0, so we decided offline to go with that.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,512 @@
...
+    func getBitsUnshifted(_ range: ClosedRange<Int>) -> Self {
+        guard range.lowerBound >= 0 else {
+            return 0
+        }
+        guard range.upperBound <= 63 else {
+            return 0
+        }
+        var value = UInt64()
+        for i in range {
+            value |= (self & (0b1 << (63 - i)))
```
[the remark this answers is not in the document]
patrickfreed: This comment still holds, but flip the shifting logic around for MSB 0.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+        // The most significant bit of the significand determines the format
+        // Encode combination, exponent, and significand.
+        if significand.hi.getBit(14) == 1 {
+            // The significand has the implicit (0b100) at the
+            // beginning of the trailing significand field
+
+            // Ensure we encode '0b11' into bits 1 and 2
+            self.value.hi.setBit(1)
+            self.value.hi.setBit(2)
+            self.value.hi |= biasedExponent << 47
```
patrickfreed, line 322: this 47 magic number would be more clear if expressed by its full derivation (i.e. `64 - combinationLength - Self.exponentLength`). Ditto for below.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+        // we can then add the significandLoDigits to the product to ensure that we have a correctly formed significand
+        var significand = UInt128.multiply(significandHiDigits, by: Self.decimalShift17Zeroes)
+        significand.lo += significandLoDigits
+
+        if significand.lo < significandLoDigits {
+            // carry over addition to hi side
+            significand.hi += 1
+        }
+
+        let biasedExponent = UInt64(exponent + Self.exponentBias).getBits(49...63)
```
patrickfreed, line 309: I think a `getLeastSignificantBits` that accepts a length could be useful here and elsewhere that have these `x...63` closed ranges. Then we could do something here like `.getLeastSignificantBits(Self.exponentLength)`

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+            }
+
+            if digits.last != 0 {
+                // We don't end in a zero and our exponent is too small
+                throw BSONError.InvalidArgumentError(message: "Underflow Error: Value too small")
+            }
+        }
+
+        guard (Self.exponentMin...Self.exponentMax).contains(exponent) else {
+            throw BSONError.InvalidArgumentError(message: "Rounding Error: Cannot round exponent \(exponent) further")
```
kmahar, line 286: this error message reflects what we are doing, but doesn't really tell the user what the problem is. I don't think the rounding we are doing is necessarily something the user would be thinking about. iiuc, the problem here is just that the value is either too large or too small, right? it might be better for the error message to just state whichever the case is (underflow/overflow)

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        guard digits.count <= Self.maxSignificandDigits else {
+            throw BSONError.InvalidArgumentError(
+                message: "Overflow Error: Value cannot exceed \(Self.maxSignificandDigits) digits"
```
kmahar, line 291: this message may be confusing because the user might have given us a number with fewer digits than this but the number of digits changed in the exponent manipulation above to be > 34. perhaps we should just have two generic underflow and overflow errors we reuse throughout here?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+        let significandLoDigits = [UInt8](digits.suffix(Self.maxSignificandDigits / 2)).decimalDigitsToUInt64()
+        let significandHiDigits = [UInt8](digits.dropLast(Self.maxSignificandDigits / 2)).decimalDigitsToUInt64()
+
+        // Multiply by one hundred quadrillion (note the seventeen zeroes)
+        // the product is the significandHiDigits "shifted" up by 17 decimal places
+        // we can then add the significandLoDigits to the product to ensure that we have a correctly formed significand
+        var significand = UInt128.multiply(significandHiDigits, by: Self.decimalShift17Zeroes)
+        significand.lo += significandLoDigits
+
+        if significand.lo < significandLoDigits {
```
kmahar, line 304: can you explain how/why we encounter this case, and why adding 1 to the significand handles it? since significand.lo is a UInt64 I don't see how we end up with a value here that is less than significandLoDigits after adding significandLoDigits.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+            significand.hi += 1
+        }
+
+        let biasedExponent = UInt64(exponent + Self.exponentBias).getBits(49...63)
+
+        self.value = UInt128()
+
+        // The most significant bit of the significand determines the format
+        // Encode combination, exponent, and significand.
+        if significand.hi.getBit(14) == 1 {
```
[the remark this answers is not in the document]
kmahar: why is the 14th bit of significand.hi the MSB? I thought the significand was the lowest 110 bits so it would be stored in `significand.hi[18...]` plus all 64 bits of `significand.lo` ?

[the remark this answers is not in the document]
nbbeeken: Patrick, you're correct that `0b100` is the pattern to look for but its also true that only 1 has to be seen in this position, however I do think it is clearer to check against 0b100

Kaitlin, The significand is indeed in the lower 110 bits when finally encoded which is what the code inside this if block will do, but upon calculation it can take up more space than that, that's where this idea of implicit bits come from, to increase the encoding's capabilities.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,483 @@
...
+}
+
+public struct BSONDecimal128: Equatable, Hashable, CustomStringConvertible {
+    // swiftlint:disable line_length
+    private static let digitsRegex = #"(?:\d+)"#
+    private static let indicatorRegex = #"(?:e|E)"#
+    private static let signRegex = #"[+-]"#
+    private static let infinityRegex = #"Infinity|Inf|infinity|inf"#
+    private static let decimalRegex = "\(digitsRegex)\\.\(digitsRegex)?|\\.?\(digitsRegex)"
+    private static let nanRegex = #"NaN"#
```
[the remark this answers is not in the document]
nbbeeken: I added the case insensitivity flag to the regex constructor call
https://github.com/mongodb/swift-bson/blob/67d53cdad391713327c06fd52fd148f919b1ede2/Sources/BSON/BSONDecimal128.swift#L204-L207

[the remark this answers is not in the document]
kmahar: whoops, missed that, perfect!

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
patrickfreed: Why can't the three most significant bits of the significand be `0b111` or `0b101`? Are we supposed to be looking at the bits of the most significant digit? That would make sense with 9 (`0b101`) being the maximum value. Looking at the wikipedia page, it states that the "most significant 4 bits of the significand" are what we should be looking at, but [this page](http://speleotrove.com/decimal/dbspec.html) refers to "the the most significant digit (MSD) of the coefficient (4 bits, which may take only the values 0 through 9)". Checking the bits of the digit doesn't make a lot of sense to me though because the digits themselves aren't stored--the binary representation of the actual number is.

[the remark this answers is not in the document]
patrickfreed: This [page](https://en.wikipedia.org/wiki/Binary_integer_decimal) goes into a little more detail, but it still doesn't mention why `0b111` cannot be the leading bits.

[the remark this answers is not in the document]
nbbeeken: The digits are what is being stored not the binary representation of them, each four bit sequence can only hold numbers 0-9. You can see this is being done in the encoding here:
https://github.com/mongodb/swift-bson/blob/cf2670520543749d88fbd5990c00414c742d2b05/Sources/BSON/BSONDecimal128.swift#L57-L61
I take the existing value and multiply by 10 (like a decimal shift, 8 becomes 80)
I then add in the next digit (like a decimal or since digits are guaranteed to be 0-9, 80 becomes 80 + value (does not exceed 89))

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        while exponent > Self.exponentMax && digits.count <= Self.maxSignificandDigits {
+            // Exponent is too large, try shifting zeros into the coefficient
+            digits.append(0)
+            exponent -= 1
+        }
+
+        while exponent < Self.exponentMin && !digits.isEmpty {
+            // Exponent is too small, try taking zeros off the coefficient
```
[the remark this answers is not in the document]
kmahar: repositioned comment looks good. still wondering about whether my interpretation of the edge case is correct though?

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
patrickfreed: Right, but you're doing numerical addition and storing it in a `UInt64`, which is binary storage. If you were storing the digits themselves, you'd be implementing the densely packed decimal (DPD) encoding. According to the spec (and your implementation), we're doing the Binary Integer Decimal (BID) encoding instead though, which just stores the significand as a regular binary integer.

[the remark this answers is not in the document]
patrickfreed: Okay, so I think I've figured out the answers to my questions, and the answers are very unsatisfying. 

The decimal format itself is defined generically in terms of a _k_ which is a multiple of 32. decimal32, decimal64, and decimal128 are just three of the first possible values for _k_. So theoretically, one could parse / generate arbitrary precision decimal\<k\>'s using the same code and just changing the input of this _k_. The result of this is that they all use the same "implicit bit" logic in the combination field. The trick that I was missing was that decimal128 _doesn't actually need the extra implicit bits_. The entirety of decimal128's range can fit normally by just encoding the exponent and significand as-is.

The calculation for that is as follows:
- the maximum significand is 34 digits (i.e. 9999...), and it can be represented using 113 bits. 
- the maximum biased exponent is 12287, which can be represented in 14 bits.
    - the two MSB of the maximum biased exponent are `0b10`. I think it is a mathematical property based on how the maximum exponents are derived from _k_ that they are never `0b11`. See the wikipedia page I linked to in a previous comment.
- The trailing significand field is 110 bits long, meaning 3 bits of the significand must be stored in the combination field (113 - 110 = 3). 
- The combination is 17 bits long though, meaning it is big enough to store all three missing bits of the significand and the entire exponent (14 + 3 = 17)
    - This is the implicit leading `0b0` case

For other formats, sometimes the implicit `0b100` will need to be used in order to fit everything. In those cases, due to some other mathematical property (I assume) the first three bits of the max significand are never `0b110` or `0b101` or `0b111`. 

The decimal128 wikipedia page actually calls out that decimal128 doesn't ever require the `0b100` case but is just following the general pattern:

>For the decimal128 format, all of these significands are out of the valid range (they begin with 2^113 > 1.038×1034), and are thus decoded as zero, but the pattern is same as decimal32 and decimal64.

So, in summary, I think we can actually simplify this code a bit and skip the parts about setting `0b11` and the implicit `0b100`, and just encode the exponent and significand directly into the combination field sequentially.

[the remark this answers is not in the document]
nbbeeken: Welp... yea I missed that! but it makes sense 128 bits is massive after all 👏  spectacular research, thank you for your efforts

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,514 @@
...
+        if didOverflow {
+            significand.hi += 1
+        }
+
+        let biasedExponent = UInt64(exponent + Self.exponentBias).getLeastSignificantBits(Self.exponentLength)
+
+        var value = UInt128()
+
+        value.hi |= biasedExponent << (63 - Self.exponentLength)
+        value.hi |= significand.hi.getLeastSignificantBits(60)
```
patrickfreed, line 324: I think this should be 48 (63 - 1 - 14) since the significand is only 113 bits (and also to make room for the exponent). I don't think it should actually matter since the higher order bits will all be 0 though, so maybe it just makes sense to do `value.hi |= significand.hi`.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,514 @@
...
+
+        if didOverflow {
+            significand.hi += 1
+        }
+
+        let biasedExponent = UInt64(exponent + Self.exponentBias).getLeastSignificantBits(Self.exponentLength)
+
+        var value = UInt128()
+
+        value.hi |= biasedExponent << (63 - Self.exponentLength)
```
patrickfreed, line 323: Could we add a comment here indicating why we can skip the conditional encoding here?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,526 @@
...
+            default: quotientI = 0
+            }
+            remainder += quotientI
+            // quotient[i] = Int(remainder / DIVISOR)
+            switch i {
+            case 0: quotient.hi = (((remainder / denominator) << 32) | quotient.hi.lower32bits)
+            case 1: quotient.hi = ((remainder / denominator).lower32bits | quotient.hi.getBitsUnshifted(0...31))
+            case 2: quotient.lo = (((remainder / denominator) << 32) | quotient.lo.lower32bits)
+            case 3: quotient.lo = ((remainder / denominator).lower32bits | quotient.lo.getBitsUnshifted(0...31))
+            default: _ = ()
```
kmahar, line 138: nit: I think the preferred way to no-op in a case is with `break`; see [here](https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html#ID139)

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,526 @@
...
+        self.hi = hi
+        self.lo = lo
+    }
+
+    internal init() {
+        self.hi = 0
+        self.lo = 0
+    }
+
+    internal static func multiply(_ left: UInt64, by right: UInt64) -> UInt128 {
```
kmahar, line 82: was just reading the UInt64 docs. could we just use this method instead to get the hi and lo values for the UInt128? https://developer.apple.com/documentation/swift/uint64/2885565-multipliedfullwidth

there are also some division related helpers you might be able to use to simplify the division logic

nbbeeken: I incorporated the multiply helper, the division one proves a bit more tricky with how the logic works around using the hi and lo portions of the number, as well as overwriting the significand with the quotient, would you be okay with keeping our existing logic?

kmahar: yup, that's fine with me! I didn't look at those ones too closely so I wasn't sure if they would help, just thought it might be worth looking into.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,500 @@
...
+
+        if didOverflow {
+            significand.hi += 1
+        }
+
+        let biasedExponent = UInt64(exponent + Self.exponentBias).getLeastSignificantBits(Self.exponentLength)
+
+        var value = UInt128()
+
+        // Normally Decimal(K) encodings would conditionally modify the combination field here
```
patrickfreed, line 301: one caveat to this is that we're supposed to treat significands that are too big as 0 according to the IEEE spec. According to the MongoDB spec, we're supposed to error in overflow/underflow scenarios though. So are we covered here?

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
patrickfreed: Actually, this is fine. This code review has gone on long enough and this is minor enough of a difference, we can leave this as-is.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,526 @@
...
+                    )
+                }
+                if digit == Character("0") && leadingZero {
+                    continue
+                }
+                if digit != Character("0") && leadingZero {
+                    // seen a non zero digit
+                    leadingZero = false
+                }
+                guard let digitValue = UInt8(String(digit)) else {
```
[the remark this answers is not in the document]
nbbeeken: I can but then I have to subtract from ascii zero (48) to get the actual digit which is the opposite of Kaitlin's comment here: https://github.com/mongodb/swift-bson/pull/6#discussion_r454052402

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,526 @@
...
+                    throw BSONError.InvalidArgumentError(
+                        message: "Syntax Error: \(digit) cannot be represented in a UInt8"
+                    )
+                }
+                digits.append(digitValue)
+            }
+        } else if decimalString.utf8.count == 1 {
```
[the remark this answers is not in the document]
nbbeeken: The regex ensure's that we won't pass in such a number, I can pull this logic out of a separate function if that makes more sense but it's not used without first validating certain things about the input string

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,526 @@
...
+            if self.isInfinity {
+                return (self.isNegative ? "-" : "") + "Infinity"
+            }
+            if self.isNaN {
+                return "NaN"
+            }
+            // The number is neither NaN nor Inf
+            // Decimal interchange floating-point formats c,2,ii
+            exponent = Int(self.value.hi.getBits(3...16))
+            significandPrefix = Int(self.value.hi.getBit(20) + 0b1000)
```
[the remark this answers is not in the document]
nbbeeken: There are forms of 0 that can use this logic to be decoded, and like these 0 values can be `0e+3` so they have an exponent to decode.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,500 @@
...
+        var representation = self.isNegative ? "-" : ""
+
+        let adjusted_exponent = exponent + (significand.count - 1)
+        if exponent > 0 || adjusted_exponent < -6 {
+            // Scientific format
+            representation += String(significand[0], radix: 10)
+            representation += significand.count > 1 ? "." : ""
+            representation += significand[1..<significand.count].map { String($0, radix: 10) }.joined(separator: "")
+            representation += "E"
+            representation += String(format: "%+d", adjusted_exponent)
```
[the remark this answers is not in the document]
nbbeeken: Yep

Sources/BSON/BSONDecimal128.swift
nbbeeken: I believe so, if I understand correctly it isn't possible to formulate a number in string form that in decimal128 encoding will encounter that case of the leading 0b11

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,505 @@
...
+                )
+            )
+        }
+
+        exponent -= Self.exponentBias
+
+        var significand128 = UInt128()
+
+        // significand prefix (implied bits) combined with removing the combination and sign fields
+        significand128.hi = UInt64((significandPrefix & 0xF) << 46) | self.value.hi.getBits(18...63)
```
patrickfreed, line 394: Sinnce the higher order bits of `value.hi` will be 0, we don't need to do `getBits(18...63)` here, right?

patrickfreed, line 394: can we replace `& 0xF` with `getLeastSignificantBits(4)`?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,505 @@
...
+             * 1e34 - 1.  If the value is greater than 1e34 - 1, the IEEE 754
+             * standard dictates that the significand is interpreted as zero.
+             */
+            isZero = true
+        }
+
+        if isZero {
+            significandDigits = ["0"]
+        } else {
+            for _ in 0...3 {
```
patrickfreed, line 418: can we replace the bounds of this for loop with `0...(Self.maxSignificandDigits / 9)`?

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,505 @@
...
+                representation += String(significandDigits)
+                return representation
+            }
+
+            var pointPosition = significandDigits.count + exponent
+
+            if pointPosition > 0 {
+                // number isn't a fraction
+                representation += String(significandDigits[0..<pointPosition])
+                significandDigits = Array(significandDigits[pointPosition...])
```
patrickfreed, line 467: this can just be `significandDigits.dropFirst(pointPosition - 1)` to avoid a new allocation.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,505 @@
...
+            } else {
+                representation += "0"
+            }
+
+            representation += "."
+
+            if pointPosition < 0 {
+                representation += String(repeating: "0", count: abs(pointPosition))
+                pointPosition += abs(pointPosition)
```
patrickfreed, line 476: I don't think `pointPosition` is used after this, so we don't need to modify it. As part of that, we can make it a `let` constant above.

Sources/BSON/BSONDecimal128.swift
nbbeeken: The high order bits can be nonzero here, the sign bit along with the exponent field is present.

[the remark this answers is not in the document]
nbbeeken: I think it's just 110 - 64, because `significand128` represents just the trailing significand field

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,505 @@
...
+    private static let exponentMax = 6111
+    /// In the spec exp_min is -6134 so we use -6176
+    private static let exponentMin = -6176
+    /// The sum of the exponent and a constant (bias) chosen to make the biased exponent’s range non-negative.
+    private static let exponentBias = 6176
+
+    /// Length in bits of the exponent field
+    private static let exponentLength: UInt64 = 14
+    /// Length in bits of the combination field
+    private static let combinationLength: UInt64 = 3
```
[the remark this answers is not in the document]
nbbeeken: Yea I thought this was confusing I think I misinterpreted a previous comment, removed.

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
nbbeeken: Indeed, it does get reported as zero we make the check in an if statement below that has a comment block referring to this case.

Sources/BSON/BSONDecimal128.swift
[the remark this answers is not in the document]
nbbeeken: I do agree it feels strange have this as a separate procedure and yet rely on the precondition of the caller, I moved it into the caller after cleaning it up, its just a for loop now.

Sources/BSON/BSONDecimal128.swift
patrickfreed: Oh right, of course.

Sources/BSON/BSONDecimal128.swift
```diff
@@ -0,0 +1,495 @@
...
+                representation += String(significandDigits)
+                return representation
+            }
+
+            let pointPosition = significandDigits.count + exponent
+
+            if pointPosition > 0 {
+                // number isn't a fraction
+                representation += String(significandDigits[0..<pointPosition])
+                significandDigits = Array(significandDigits.dropFirst(pointPosition))
```
patrickfreed, line 458: this also needs to add 0's for the difference between exponent and number of digits. Currently `Decimal128("20e2").toString() is giving "20".

nbbeeken: Nice catch, the regex actually didn't make the sign on the exponent optional so it was encoding `0` as the exponent.

PR #7, opened 2020-06-03
Add Workflow Scripts

nbbeeken:
Here be the Makefile, add json files to xcode for testing in xcode env, the linuxmain stencil for testing and a git hook 
We can add to the readme or dev guide that one can set up the git hook with `make install_hook` or we can drop this if we don't wanna be prescriptive here

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ project:
 
 linuxmain:
 	$(SOURCERY) --sources Tests/ --exclude-sources Tests/DisabledTests/ --templates Tests/LinuxMain.stencil --output Tests/LinuxMain.swift
+	# Tail the stenciled file to remove the sourcery inserted header, so you don't get version # conflicts in CI
 	@tail -n +5 Tests/LinuxMain.swift > Tests/LinuxMain.swift.tmp
 	@mv Tests/LinuxMain.swift.tmp Tests/LinuxMain.swift
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,62 @@
+# if provided, FILTER is used as the --filter argument to `swift test`.
+ifdef FILTER
+	FILTERARG = --filter $(FILTER)
+else
+	FILTERARG =
+endif
+
+ifdef DOCSVERSION
+	DOCSARG = --module-version $(DOCSVERSION)
+else
+	DOCSARG =
+endif
+
+# if no value provided assume sourcery is in the user's PATH
+SOURCERY ?= sourcery
+
+define check_for_gem
+	gem list $(1) -i > /dev/null || gem install $(1) || { echo "ERROR: Failed to locate or install the ruby gem $(1); please install yourself with 'gem install $(1)' (you may need to use sudo)"; exit 1; }
+endef
+
+all:
+	swift build
+
+# project generates the .xcodeproj, and then modifies it to add
+# spec .json files to the project
+project:
+	swift package generate-xcodeproj
+	@$(call check_for_gem,xcodeproj)
+	ruby etc/add_json_files.rb
+
+linuxmain:
+	$(SOURCERY) --sources Tests/ --exclude-sources Tests/DisabledTests/ --templates Tests/LinuxMain.stencil --output Tests/LinuxMain.swift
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,62 @@
...
+# project generates the .xcodeproj, and then modifies it to add
+# spec .json files to the project
+project:
+	swift package generate-xcodeproj
+	@$(call check_for_gem,xcodeproj)
+	ruby etc/add_json_files.rb
+
+linuxmain:
+	$(SOURCERY) --sources Tests/ --exclude-sources Tests/DisabledTests/ --templates Tests/LinuxMain.stencil --output Tests/LinuxMain.swift
+	@tail -n +5 Tests/LinuxMain.swift > Tests/LinuxMain.swift.tmp
```
kmahar, line 33: what are we doing with tail and the temp file here? think a comment would be helpful for future readers

Tests/LinuxMain.stencil
```diff
@@ -0,0 +1,22 @@
+@testable import BSONTests
+@testable import MongoSwiftTests
```
kmahar, line 2: I'd think we need to remove this and the next line in order for things to compile, given that those modules don't exist in this library?

Makefile
```diff
@@ -0,0 +1,62 @@
+# if provided, FILTER is used as the --filter argument to `swift test`.
+ifdef FILTER
+	FILTERARG = --filter $(FILTER)
+else
+	FILTERARG =
+endif
+
+ifdef DOCSVERSION
```
kmahar, line 8: I think we can get rid of this - it's an artifact of how we used to generate the driver docs, and should be taken out of the driver's makefile too.

Makefile
```diff
@@ -0,0 +1,62 @@
...
+	@mv Tests/LinuxMain.swift.tmp Tests/LinuxMain.swift
+
+test:
+	swift test -v $(FILTERARG)
+
+test-pretty:
+	@$(call check_for_gem,xcpretty)
+	@set -o pipefail && swift test $(FILTERARG) 2>&1 | xcpretty
+
+SOURCES := $(shell find . -name "*.swift" | grep -v ".build")
```
kmahar, line 43: if I'm not mistaken, swiftformat and swiftlint both already handle excluding the build directory in their config files, so I think we can get rid of SOURCES, just doing `swiftfomat .` and `swiftlint autocorrect`

etc/pre-commit.sh
```diff
@@ -0,0 +1,30 @@
+#! /usr/bin/env bash
```
kmahar, line 1: I think it's a fine idea to have a pre-commit hook around that people can install if they want to; I've never been a user of them myself but it's nice to have the option.
that said, I think for maintainability / any future customization other users might want to do, it would be useful to add more comments in here explaining how it works. for example, as a casual reader I'm not totally sure why FILES_CHANGED is being determined in that way with the filter and all

Makefile
nbbeeken: Silly trick here that I think we just want better CI to do, but it removes the Sourcery version text so you don't have to manually change it 😅

kmahar: sgtm, can we add a comment so we remember that in the future?

Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+test:
+	swift test -v $(FILTERARG)
+
+test-pretty:
+	@$(call check_for_gem,xcpretty)
+	@set -o pipefail && swift test $(FILTERARG) 2>&1 | xcpretty
+
+lint:
+	@swiftformat .
```
kmahar, line 38: do we wnat to suppress all of the linting output with the `@`s (if I remember correctly what that does)? seems like it would be helpful to see what went wrong printed out? (same re test-pretty)

nbbeeken: It suppresses printing out the command not the output, I think make dumps the full command by default cus its usually a big constructed gcc call or w.e. I figured here there's nothing being constructed, cleans up your output a bit, its entirely opinionated and loosely held

kmahar: oh right, yes that seems fine to me then!

PR #8, opened 2020-06-04
SWIFT-868: Document Iterator findByteRange

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-868

The initial implementation was over optimized and has been simplified, base branch is the int PR until that gets in to keep the diff readable.


Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -28,37 +28,128 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
 
-        guard let type = BSONType(rawValue: typeByte), type != .invalid else {
-            guard self.buffer.readableBytes != 0 else {
-                // Iteration exhuasted!
+    /**
+     * Find the starting index and length of a BSON Element
+     * - Parameter for: the key used to locate the element
+     */
+    internal func findByteRange(for key: String) -> (startIndex: Int, length: Int)? {
+        let key = [UInt8](key.utf8 + [0])
```
kmahar, line 49: I appreciate the effort to make this more efficient, but I think this all falls into the category of optimization work we chose to omit from this project and tackle all together later on to keep the scope manageable.

as agreed upon in the design the implementation of this method is meant to be naive/straightforward:
> It can be implemented by tracking the readerIndex and repeatedly calling next() until the desired key is found. The return value is then (readerIndex before key is found, readerIndex after key is found).

it's likely that in future optimization work we'll do something like e.g. maintain a table of byte offsets that would give you `findByteRange` in O(1) and make the logic here unneeded, but without a whole design doc on that the best approach is TBD. in the interest of getting the project done quickly, let's try to keep the focus on simple code that works

nbbeeken: My honest mistake, I overlooked that and over coded here, I'll put this in that bson optimizations google doc we have maybe it would get some use, I agree for the first run on this simplicity is ideal

PR #10, opened 2020-06-05
SWIFT-845: Add support for Array, Bool, String, Double, Date

nbbeeken:
_DRAFT_
https://jira.mongodb.org/browse/SWIFT-845

### DIFF
--- a/Sources/BSON/Array+BSONValue.swift
+++ b/Sources/BSON/Array+BSONValue.swift
@@ -1,0 +1,23 @@
+import NIO
+
+/// An extension of `Array` to represent the BSON array type.
+extension Array: BSONValue where Element == BSON {
+    internal static var bsonType: BSONType { .array }
+
+    internal var bson: BSON { .array(self) }
+
+    internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
+        guard let doc = try BSONDocument.read(from: &buffer).documentValue else {
+            throw BSONError.InternalError(message: "BSON Array cannot be read, failed to get documentValue")
+        }
+        return .array(doc.values)
+    }
+
+    internal func write(to buffer: inout ByteBuffer) {
+        var array = BSONDocument()
+        for (index, value) in self.enumerated() {
+            array[String(index)] = value
+        }
+        array.write(to: &buffer)
+    }
+}
--- a/Sources/BSON/BSONDocument.swift
+++ b/Sources/BSON/BSONDocument.swift
@@ -297,12 +297,19 @@ public struct BSONDocument {
             // Implicitly validate with iterator
             do {
                 while let (_, value) = try iter.nextThrowing() {
-                    if let doc = value.documentValue {
+                    switch value {
+                    case let .document(doc):
                         try doc.storage.validate()
+                    case let .array(array):
+                        for item in array {
+                            if let doc = item.documentValue {
+                                try doc.storage.validate()
+                            }
+                        }
+                    default:
+                        _ = ()
                     }
                 }
-            } catch {
-                throw BSONError.InvalidArgumentError(message: "Validation failed: \(error)")
             }
         }
     }
--- a/Sources/BSON/BSONDocument.swift
+++ b/Sources/BSON/BSONDocument.swift
@@ -307,7 +307,7 @@ public struct BSONDocument {
                             }
                         }
                     default:
-                        _ = ()
+                        continue
                     }
                 }
             }
--- a/Sources/BSON/Bool+BSONValue.swift
+++ b/Sources/BSON/Bool+BSONValue.swift
@@ -1,11 +1,11 @@
 import NIO
 
 extension Bool: BSONValue {
-    static var bsonType: BSONType { .bool }
+    internal static var bsonType: BSONType { .bool }
 
-    var bson: BSON { .bool(self) }
+    internal var bson: BSON { .bool(self) }
 
-    static func read(from buffer: inout ByteBuffer) throws -> BSON {
+    internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
         guard let value = buffer.readInteger(as: UInt8.self) else {
             throw BSONError.InternalError(message: "Could not read Bool")
         }
@@ -15,7 +15,7 @@ extension Bool: BSONValue {
         return .bool(value == 0 ? false : true)
     }
 
-    func write(to buffer: inout ByteBuffer) {
+    internal func write(to buffer: inout ByteBuffer) {
         buffer.writeBytes([self ? 1 : 0])
     }
 }
--- a/Sources/BSON/Bool+BSONValue.swift
+++ b/Sources/BSON/Bool+BSONValue.swift
@@ -1,0 +1,21 @@
+import NIO
+
+extension Bool: BSONValue {
+    static var bsonType: BSONType { .bool }
+
+    var bson: BSON { .bool(self) }
+
+    static func read(from buffer: inout ByteBuffer) throws -> BSON {
+        guard let value = buffer.readInteger(as: UInt8.self) else {
+            throw BSONError.InternalError(message: "Could not read Bool")
+        }
+        guard value == 0 || value == 1 else {
+            throw BSONError.InternalError(message: "Bool must be 0 or 1, found:\(value)")
+        }
+        return .bool(value == 0 ? false : true)
+    }
+
+    func write(to buffer: inout ByteBuffer) {
+        buffer.writeBytes([self ? 1 : 0])
+    }
+}
--- a/Sources/BSON/ByteBuffer+BSON.swift
+++ b/Sources/BSON/ByteBuffer+BSON.swift
@@ -1,30 +1,50 @@
 import NIO
 
 extension ByteBuffer {
-    /// Write null terminated string into this ByteBuffer using UTF-8 encoding,
-    /// moving the writer index forward by the byte length of string + 1 (for null terminator).
+    /// Write null terminated UTF-8 string to ByteBuffer starting at writerIndex
     @discardableResult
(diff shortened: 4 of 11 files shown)


Sources/BSON/Date+BSONValue.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    /// Initializes a new `Date` representing the instance `msSinceEpoch` milliseconds
+    /// since the Unix epoch.
+    internal init(msSinceEpoch: Int64) {
+        self.init(timeIntervalSince1970: TimeInterval(msSinceEpoch) / 1000.0)
+    }
+
+    static func read(from buffer: inout ByteBuffer) throws -> BSON {
+        guard let ms = buffer.readInteger(endianness: .little, as: Int64.self) else {
+            throw BSONError.InternalError(message: "Unable to read int")
```
kmahar, line 20: maybe add a little more context to the message e.g. "while reading Date"?

Sources/BSON/Double+BSONValue.swift
```diff
@@ -0,0 +1,22 @@
...
+
+    var bson: BSON {
+        .double(self)
+    }
+
+    static func read(from buffer: inout ByteBuffer) throws -> BSON {
+        guard let data = buffer.readBytes(length: 8) else {
+            throw BSONError.InternalError(message: "Cannot read 8 bytes")
+        }
+        let value = data.withUnsafeBytes { $0.load(as: Double.self) }
```
kmahar, line 14: does this ever fail? / what happens if it can't read a double?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -105,7 +105,13 @@ final class BSONCorpusTests: BSONTestCase {
         let INCLUDED_CORPUS_TESTS = [
             "Int32 type",
             "Int64 type",
-            "Decimal128"
+            "Decimal128",
+            "Boolean",
+            "DateTime",
+            "Array",
+            "Document type (sub-documents)",
```
kmahar, line 112: I guess this one already worked but we just weren't including it?

nbbeeken: The sub-document test's need string support so we can only test this now

Sources/BSON/Double+BSONValue.swift
nbbeeken: I guess not, according to my research obtaining $0 which is a pointer to the underlying memory can't fail, we know its 8 bytes because of the readBytes call. The load is no more involved than a C-like cast, so at that point if the values are not in floating point format it will just be a garbled number.
Maybe I can call `Double(exactly: value)` and error check? Is that too much validation?

Sources/BSON/Array+BSONValue.swift
```diff
@@ -0,0 +1,23 @@
+import NIO
+
+/// An extension of `Array` to represent the BSON array type.
+extension Array: BSONValue where Element == BSON {
+    static var bsonType: BSONType { .array }
```
patrickfreed, line 5: all these need explicit acl's. It is a bit redundant because `BSONValue` is `internal` itself, but it's nice to have that information included here as well.

Sources/BSON/BSONDocument.swift
```diff
@@ -283,6 +283,13 @@ public struct BSONDocument {
                 if let doc = value.documentValue {
                     try BSONDocument.validate(doc.buffer)
                 }
+                if let array = value.arrayValue {
```
patrickfreed, line 286: since `BSON` is an enum, this looks like a good situation to use the `switch` syntax instead of repeated `if let`s

Sources/BSON/Double+BSONValue.swift
patrickfreed: From this [answer](https://stackoverflow.com/a/38024025) on stackoverflow, it looks like it's safer to copy the bytes to an already instantiated `Double` rather than loading one directly.

Tests/BSONTests/BSONDocumentIteratorTests.swift
```diff
@@ -4,25 +4,25 @@ import Nimble
 import XCTest
 
 final class DocumentIteratorTests: BSONTestCase {
```
patrickfreed, line 6: I think the indentation is off in this file

Tests/BSONTests/BSONTests.swift
```diff
@@ -19,6 +19,15 @@ open class BSONTestCase: XCTestCase {
         }
         return path
     }
+
+    struct TestError: Error {
+        let message: String
+    }
+
+    func fail(_ message: String = "This should not happen") -> Never {
```
patrickfreed, line 27: I don't think we would ever want to use this. `fatalError` will crash the entire test runner and prevent future tests from compelting.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -114,6 +121,9 @@ final class BSONCorpusTests: BSONTestCase {
                         XCTFail("Unable to interpret canonical_bson as Data")
                         return
                     }
+                    let description = "\(testFile.description)-\(test.description)"
+
+                    print("VALID: \(description)")
```
patrickfreed, line 126: this debug print and the one below should be removed

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -1,30 +1,52 @@
...
             }
-            bytes += b
         }
         throw BSONError.InternalError(message: "Failed to read CString, possibly missing null terminator?")
     }
+
+    /// Get a BSONType byte from self returns .invalid for unknown types.
+    internal func getBSONType(at position: Int) -> BSONType {
+        let typeByte = self.getInteger(at: position, as: UInt8.self) ?? BSONType.invalid.rawValue
+        guard let type = BSONType(rawValue: typeByte), type != .invalid else {
```
patrickfreed, line 46: this can be simplified a bit by using [`Optional.flatMap`](https://developer.apple.com/documentation/swift/optional/1540500-flatmap): 
```swift
self.getInteger(at: position).flatMap(BSONType.init(rawValue:)) ?? .invalid
```
This is the one-liner approach, but it can be made more verbose by adding a `guard` and a small closure if you think it would be clearer:
```swift
guard let bsonType = self.getInteger(at: position).flatMap({ BSONType.init(rawValue: $0) }) else {
    return .invalid
}
return bsonType
```

Sources/BSON/BSONDocument.swift
```diff
@@ -297,12 +297,19 @@ public struct BSONDocument {
...
+                    case let .document(doc):
                         try doc.storage.validate()
+                    case let .array(array):
+                        for item in array {
+                            if let doc = item.documentValue {
+                                try doc.storage.validate()
+                            }
+                        }
+                    default:
+                        _ = ()
```
patrickfreed, line 310: this can just be `continue` to skip to the next iteration

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -1,30 +1,50 @@
...
+    internal func getCString(at offset: Int) throws -> String {
+        let key = try getBSONKey(at: offset).dropLast()
+        guard let string = String(bytes: key, encoding: .utf8) else {
+            throw BSONError.InternalError(message: "Failed to decode BSONKey as UTF8: \(key)")
+        }
+        return string
+    }
+
+    /// Returns the C String key including the null byte, for ease of comparison and byte counting
+    internal func getBSONKey(at offset: Int) throws -> [UInt8] {
```
patrickfreed, line 28: The docstring mentions this is needed for "ease of comparison and byte counting", but I don't actually see where either of those are occurring. Does this need to be a separate method? It only seems to be used in `readCString`.

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -1,30 +1,50 @@
...
+            } else {
+                throw BSONError.InternalError(message: "Failed to read CString, unable to read byte from \(self)")
             }
-            bytes += b
         }
         throw BSONError.InternalError(message: "Failed to read CString, possibly missing null terminator?")
     }
+
+    /// Get a BSONType byte from self returns .invalid for unknown types.
+    internal func getBSONType(at position: Int) -> BSONType {
```
patrickfreed, line 44: this also doesn't seem to be used anywhere, do we need to keep it around? (I should have asked this before I commented on its contents...)

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -3,6 +3,7 @@ import Foundation
 import Nimble
 import XCTest
 
+@available(OSX 10.13, *)
```
patrickfreed, line 6: is there a reason this is necessary?

Sources/BSON/Bool+BSONValue.swift
```diff
@@ -0,0 +1,21 @@
+import NIO
+
+extension Bool: BSONValue {
+    static var bsonType: BSONType { .bool }
```
patrickfreed, line 4: these need explicit acl's also. ditto for other `BSONValues` in this PR. It seems swiftlint doesn't properly check for `explicit_acl` in extensions unfortunately.

Sources/BSON/Date+BSONValue.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import NIO
+
+extension Date: BSONValue {
+    static var bsonType: BSONType { .datetime }
```
patrickfreed, line 5: this extension and and the one for `String` need some acls

nbbeeken: I obviously rely too much on the linter! thanks for catching these

PR #11, opened 2020-06-11
SWIFT-868,SWIFT-867,SWIFT-866: Mutable Valid BSONDocument

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-867
https://jira.mongodb.org/browse/SWIFT-868
https://jira.mongodb.org/browse/SWIFT-866

This will make BSONDocument's mutable, this is the work of three tickets of implementing the ability to search and splice a byte buffer to provide Create Update and Delete actions on a document. Since I rolled back the significantly more complex document iterator, the work compacted quite a bit so document validation is also in here. And validation is being tested by the corpus runner

### DIFF
--- a/Sources/BSON/BSONDocument.swift
+++ b/Sources/BSON/BSONDocument.swift
@@ -14,11 +14,11 @@ public struct BSONDocument {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BSON)
 
-    private var _buffer: ByteBuffer
+    internal var storage: BSONDocumentStorage
 
     internal var byteLength: Int {
         get {
-            guard let byteLength = self._buffer.getInteger(at: 0, endianness: .little, as: Int32.self) else {
+            guard let byteLength = self.storage.buffer.getInteger(at: 0, endianness: .little, as: Int32.self) else {
                 fatalError("Cannot read byteLength of BSON from buffer")
             }
             return Int(byteLength)
@@ -27,7 +27,7 @@ public struct BSONDocument {
             guard newValue <= Int32.max else {
                 fatalError("Cannot cast \(newValue) down to Int32")
             }
-            self._buffer.setInteger(Int32(newValue), at: 0, endianness: .little, as: Int32.self)
+            self.storage.buffer.setInteger(Int32(newValue), at: 0, endianness: .little, as: Int32.self)
         }
     }
 
@@ -44,38 +44,38 @@ public struct BSONDocument {
             fatalError("Dictionary \(keyValuePairs) contains duplicate keys")
         }
 
-        self._buffer = BSON_ALLOCATOR.buffer(capacity: 0)
+        self.storage = BSONDocumentStorage(BSON_ALLOCATOR.buffer(capacity: 0))
 
         guard !self.keySet.isEmpty else {
             self = BSONDocument()
             return
         }
 
-        let start = self._buffer.writerIndex
+        let start = self.storage.buffer.writerIndex
 
         // reserve space for our byteLength that will be calculated
-        self._buffer.writeInteger(0, endianness: .little, as: Int32.self)
+        self.storage.buffer.writeInteger(0, endianness: .little, as: Int32.self)
 
         for (key, value) in keyValuePairs {
-            BSONDocument.append(key: key, value: value, to: &self._buffer)
+            BSONDocumentStorage.append(key: key, value: value, to: &self.storage.buffer)
         }
         // BSON null terminator
-        self._buffer.writeInteger(0, as: UInt8.self)
+        self.storage.buffer.writeInteger(0, as: UInt8.self)
 
-        guard let byteLength = Int32(exactly: self._buffer.writerIndex - start) else {
-            fatalError("Data is \(self._buffer.writerIndex - start) bytes, "
+        guard let byteLength = Int32(exactly: self.storage.buffer.writerIndex - start) else {
+            fatalError("Data is \(self.storage.buffer.writerIndex - start) bytes, "
                 + "but maximum allowed BSON document size is \(Int32.max) bytes")
         }
         // Set byteLength in reserved space
-        self._buffer.setInteger(byteLength, at: 0, endianness: .little, as: Int32.self)
+        self.storage.buffer.setInteger(byteLength, at: 0, endianness: .little, as: Int32.self)
     }
 
     /// Initializes a new, empty `BSONDocument`.
     public init() {
         self.keySet = Set()
-        self._buffer = BSON_ALLOCATOR.buffer(capacity: 0)
-        self._buffer.writeInteger(5, endianness: .little, as: Int32.self)
-        self._buffer.writeBytes([0])
+        self.storage = BSONDocumentStorage(BSON_ALLOCATOR.buffer(capacity: 0))
+        self.storage.buffer.writeInteger(5, endianness: .little, as: Int32.self)
+        self.storage.buffer.writeBytes([0])
     }
 
     /**
@@ -102,13 +102,14 @@ public struct BSONDocument {
      * - SeeAlso: http://bsonspec.org/
      */
     public init(fromBSON bson: ByteBuffer) throws {
-        try BSONDocument.validate(bson)
-        self = BSONDocument(fromUnsafeBSON: bson)
+        let storage = BSONDocumentStorage(bson)
+        try storage.validate()
+        self = BSONDocument(fromUnsafeBSON: storage)
     }
 
-    internal init(fromUnsafeBSON bson: ByteBuffer) {
+    internal init(fromUnsafeBSON storage: BSONDocumentStorage) {
         self.keySet = Set()
-        self._buffer = bson
+        self.storage = storage
         for (key, _) in self {
             self.keySet.insert(key)
         }
@@ -126,10 +127,10 @@ public struct BSONDocument {
     /// A copy of the `ByteBuffer` backing this document, containing raw BSON data. As `ByteBuffer`s implement
     /// copy-on-write, this copy will share byte storage with this document until either the document or the returned
     /// buffer is mutated.
-    public var buffer: ByteBuffer { self._buffer }
+    public var buffer: ByteBuffer { self.storage.buffer }
 
     /// Returns a `Data` containing a copy of the raw BSON data backing this document.
-    public func toData() -> Data { Data(self._buffer.readableBytesView) }
+    public func toData() -> Data { Data(self.storage.buffer.readableBytesView) }
 
     /// Returns a `Boolean` indicating whether this `BSONDocument` contains the provided key.
     public func hasKey(_ key: String) -> Bool { self.keySet.contains(key) }
@@ -155,7 +156,7 @@ public struct BSONDocument {
         set {
             // The only time this would crash is document too big error
             // swiftlint:disable force_try
-            try! self.set(key: key, value: newValue)
+            try! self.set(key: key, to: newValue)
         }
     }
 
@@ -194,7 +195,7 @@ public struct BSONDocument {
      * Sets a BSON element with the corresponding key
      * if element.value is nil the element is deleted from the BSON
      */
-    internal mutating func set(key: String, value: BSON?) throws {
+    internal mutating func set(key: String, to value: BSON?) throws {
         if !self.keySet.contains(key) {
             guard let value = value else {
                 // no-op: key does not exist and the value is nil
@@ -202,14 +203,14 @@ public struct BSONDocument {
             }
             // appending new key
             self.keySet.insert(key)
-            self._buffer.moveWriterIndex(to: self.byteLength - 1) // setup to overwrite null terminator
-            let size = BSONDocument.append(key: key, value: value, to: &self._buffer)
-            self._buffer.writeInteger(0, endianness: .little, as: UInt8.self) // add back in our null terminator
+            self.storage.buffer.moveWriterIndex(to: self.byteLength - 1) // setup to overwrite null terminator
+            let size = BSONDocumentStorage.append(key: key, value: value, to: &self.storage.buffer)
+            self.storage.buffer.writeInteger(0, endianness: .little, as: UInt8.self) // add back in our null terminator
             self.byteLength += size
             return
         }
 
-        var iter = BSONDocumentIterator(over: self._buffer)
+        var iter = BSONDocumentIterator(over: self.storage.buffer)
 
         guard let range = iter.findByteRange(for: key) else {
             throw BSONError.InternalError(message: "Cannot find \(key) to delete")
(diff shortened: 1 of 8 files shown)


Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -28,13 +28,25 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+     */
+    internal mutating func _next() throws -> BSONDocument.KeyValuePair? {
         guard self.buffer.readableBytes != 0 else {
             // Iteration has been exhausted
             return nil
         }
 
         guard let typeByte = self.buffer.readInteger(as: UInt8.self) else {
-            fatalError("BSONDocumentIterator Failed: Cannot read from \(self.buffer)")
+            throw BSONError.InternalError(message: "BSONDocumentIterator Failed: Cannot read from \(self.buffer)")
```
kmahar, line 49: I think all of the error message in this method might be more helpful if we could include the byte offset at which we found invalid BSON

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -43,31 +55,39 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
         }
+        return (key: key, value: bson)
     }
 
     /// Finds the key in the underlying buffer, and returns the [startIndex, endIndex) containing the corresponding
     /// element.
-    internal func findByteRange(for key: String) -> (startIndex: Int, endIndex: Int) {
-        fatalError("Unimplemented")
+    internal func findByteRange(for searchKey: String) -> (startIndex: Int, endIndex: Int)? {
+        var iter = BSONDocumentIterator(over: self.buffer)
```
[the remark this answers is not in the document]
nbbeeken: Since the buffer has to be mutated to track the changes in the readerIndex and currently the iterator's buffer is private I think its fine to leave in here. I've updated to use `self`

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +185,107 @@ public struct BSONDocument {
...
+            self.keySet.insert(key)
+            self._buffer.moveWriterIndex(to: self.byteLength - 1) // setup to overwrite null terminator
+            let size = BSONDocument.append(element: (key, value), to: &self._buffer)
+            self._buffer.writeInteger(0, endianness: .little, as: UInt8.self) // add back in our null terminator
+            self.byteLength += size
+            return
+        }
+
+        guard value != nil || self.keySet.contains(key) else {
+            // no-op: trying to delete a key that doesn't exist
```
kmahar, line 207: can we add a test for this behavior? 

also just a nit but I think the logic might be clearer if we can combine this with the check above rather than checking whether the key is in the set and the value is non-nil twice, maybe something like

```
if keyset doesn't contain key:
    guard let value = value else return early (no op case)
    add new key
    return
```

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +185,107 @@ public struct BSONDocument {
...
+
+        newBuffer.writeBytes(prefix)
+
+        var newSize = self.byteLength - (end - start)
+        if let value = value {
+            // Overwriting
+            let size = BSONDocument.append(element: (key, value), to: &newBuffer)
+            newSize += size
+
+            guard newSize != BSON_MAX_SIZE else {
```
kmahar, line 237: probably also need to handle newSize > BSON_MAX_SIZE?

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +185,107 @@ public struct BSONDocument {
...
+            // Deleting
+            self.keySet.remove(key)
+        }
+
+        newBuffer.writeBytes(suffix)
+
+        self._buffer = newBuffer
+        self.byteLength = newSize
+        guard self.byteLength == self._buffer.readableBytes else {
+            fatalError("I think the bson is \(self.byteLength) but I can only read \(self._buffer.readableBytes)")
```
kmahar, line 250: I think we generally try to avoid speaking in the first person in error messages, can we rephrase?

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +185,107 @@ public struct BSONDocument {
...
+        guard let byteLength = bson.getInteger(at: 0, endianness: .little, as: Int32.self) else {
+            throw BSONError.InvalidArgumentError(message: "Validation Failed: Cannot read byteLength")
+        }
+
+        guard byteLength >= BSON_MIN_SIZE && byteLength <= BSON_MAX_SIZE else {
+            throw BSONError.InvalidArgumentError(message: "Validation Failed: BSON cannot be \(byteLength) bytes long")
+        }
+
+        guard byteLength == bson.writerIndex else {
+            throw BSONError.InvalidArgumentError(message: "Validation Failed: Cannot read \(byteLength) from bson")
```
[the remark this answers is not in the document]
nbbeeken: `byteLength == bson.readableBytes` is a better test I think, But I still add some kind of comment to the effect of this expectation

Sources/BSON/BSONDocument.swift
nbbeeken: Silly debug phrasing 😅  nice catch

Sources/BSON/BSONDocument.swift
```diff
@@ -88,7 +93,8 @@ public struct BSONDocument {
      * Initializes a new `BSONDocument` from the provided BSON data.
      *
      * - Throws:
-     *   - `InvalidArgumentError` if the data passed is invalid BSON
+     *   - `InvalidArgumentError` if the data passed is invalid BSON,
+     *      ByteBuffer must have `readableBytes` equal to the BSON's leading size indicator
```
kmahar, line 97: I think maybe we should put this at the top after the first sentence like

Initializes a new `BSONDocument` from the provided BSON data. The buffer must have `readableBytes` equal to the BSON's leading size indicator.

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +186,107 @@ public struct BSONDocument {
...
         get { self[member] }
-        set { fatalError("Unimplemented") }
+        set { self[member] = newValue }
+    }
+
+    /**
+     * Sets a BSON element with the corresponding key
+     * if element.value is nil the element is deleted from the BSON
+     */
+    internal mutating func set(key: String, of value: BSON?) throws {
```
kmahar, line 196: `of` -> `to`?

Sources/BSON/BSONDocument.swift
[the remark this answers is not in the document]
kmahar: sgtm, let's just update the error string then

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +186,107 @@ public struct BSONDocument {
...
+        if !self.keySet.contains(key) {
+            guard let value = value else {
```
kmahar, line 198: new logic looks good but i think we still need to test this behavior

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +186,107 @@ public struct BSONDocument {
...
+        self.byteLength = newSize
+        guard self.byteLength == self._buffer.readableBytes else {
+            fatalError("BSONDocument's encoded byte length is \(self.byteLength) however the" +
+                "buffer has \(self._buffer.readableBytes) readable bytes")
+        }
+    }
+
+    /// Appends element to underlying BSON bytes, returns the size of the element appended: type + key + value
+    @discardableResult
+    internal static func append(key: String, of value: BSON, to buffer: inout ByteBuffer) -> Int {
```
kmahar, line 257: I think "of" is also kind of funny here, as at the call site it is like "append key a of 1 to buffer". but I will leave you and @patrickfreed to bikeshed that 🙂

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -43,31 +58,48 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+            )
         }
+        return (key: key, value: bson)
     }
 
     /// Finds the key in the underlying buffer, and returns the [startIndex, endIndex) containing the corresponding
     /// element.
-    internal func findByteRange(for key: String) -> (startIndex: Int, endIndex: Int) {
-        fatalError("Unimplemented")
+    internal mutating func findByteRange(for searchKey: String) -> (startIndex: Int, endIndex: Int)? {
```
patrickfreed, line 83: Instead of our own tuple type, we can use the stdlib [`Range<T>`](https://developer.apple.com/documentation/swift/range) and return a `Range<Int>?`.

Sources/BSON/BSONError.swift
```diff
@@ -40,3 +41,27 @@ public enum BSONError {
         }
     }
 }
+
+internal func BSONIterationError(
```
patrickfreed, line 45: this could use a docstring explaining what it is doing.

Tests/BSONTests/BSONDocumentIteratorTests.swift
```diff
@@ -0,0 +1,28 @@
...
+        let d: BSONDocument = [:]
+        var iter = d.makeIterator()
+        let range = iter.findByteRange(for: "item")
+        expect(range).to(beNil())
+    }
+
+    func testFindByteRangeItemsInt32() {
+        let d: BSONDocument = ["item0": .int32(32), "item1": .int32(32)]
+        var iter = d.makeIterator()
+        let range = iter.findByteRange(for: "item1")!
```
patrickfreed, line 17: If this unwrap fails, it will crash the entire test runner. I think it would be better to first verify that it is not nil via an `expect(...).toNot(beNil())` or via a `guard let ... else { XCTFail("reason") }` to prevent that.

Sources/BSON/BSONDocument.swift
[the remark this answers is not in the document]
nbbeeken: Maybe the BSONDocument is just getting too large and I could pull this and set out into an extension but I worry about the storage idea being a frustrating debugging experience, I've seen this pattern done for ByteBuffer and Data where they use a "storage" dataclass to wrap a pointer it's just adding yet another layer for those who need to peer into their BSON binary

Sources/BSON/BSONDocument.swift
```diff
@@ -158,7 +170,7 @@ public struct BSONDocument {
      *  ```
      */
     public subscript(key: String, default defaultValue: @autoclosure () -> BSON) -> BSON {
-        fatalError("Unimplemented")
+        self[key] ?? defaultValue()
```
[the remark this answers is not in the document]
nbbeeken: Nope! it calls the subscript function a few lines up to keep us DRY ☔

Sources/BSON/BSONDocument.swift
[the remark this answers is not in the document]
patrickfreed: I think this still needs to be addressed

Sources/BSON/BSONDocument.swift
```diff
@@ -174,12 +187,108 @@ public struct BSONDocument {
...
         get { self[member] }
-        set { fatalError("Unimplemented") }
+        set { self[member] = newValue }
+    }
+
+    /**
+     * Sets a BSON element with the corresponding key
+     * if element.value is nil the element is deleted from the BSON
+     */
+    internal mutating func set(key: String, value: BSON?) throws {
```
patrickfreed, line 197: I think this one `to` actually makes more sense as a label for the value

Sources/BSON/BSONDocument.swift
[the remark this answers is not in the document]
patrickfreed: >Maybe the BSONDocument is just getting too large and I could pull this and set out into an extension

It's not necessarily just about stuff on `BSONDocument` though. Some related functionality is also added as extensions to `ByteBuffer` in the later PRs, so the logic for operating on BSON types with `ByteBuffer` is spread around the codebase in ways that aren't consistently discoverable. By introducing a new type that operates over the buffer, we can organize all this functionality into a single, easily discoverable place. 

>I worry about the storage idea being a frustrating debugging experience, I've seen this pattern done for ByteBuffer and Data where they use a "storage" dataclass to wrap a pointer it's just adding yet another layer for those who need to peer into their BSON binary

As far as causing debugging frustration, I guess I'm not really seeing what the concern is. Keep in mind this `DocumentStorage` would be a `struct`, so it wouldn't change the reference counting / CoW semantics of `ByteBuffer`. For me personally, encapsulating all this behavior in a single place would make debugging easier, since I could more easily enumerate all the places that touch the buffer.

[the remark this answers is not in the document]
nbbeeken: Alright yea I see the benefit there, I'll move the BSON modifying functions append and set as well as validate to a storage type, but I have to leave the extensions to ByteBuffer there since they're needed in reading and writing types and the BSONValue protocol operates on ByteBuffer. 
I was too vague about the debug frustration 😅  it's literally in the xcode UI there'd be 2 more layers of drop downs to see the internals of a BSON document. But that's a very specific developer pain so not too relevant

Sources/BSON/BSONDocument.swift
```diff
@@ -150,11 +153,9 @@ public struct BSONDocument {
             return nil
         }
         set {
-            do {
-                try self.set(key: key, of: newValue)
-            } catch {
-                fatalError("\(error)")
-            }
+            // The only time this would crash is document too big error
```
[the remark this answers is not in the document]
nbbeeken: This would change from the fatalError behavior in the driver. It's not exactly something one could rely on (like API wise) but going from crash to silence doesn't seem right here, thoughts?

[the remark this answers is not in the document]
patrickfreed: Oh if that is the current behavior of the driver, then we should preserve it for now. I filed [SWIFT-892](https://jira.mongodb.org/browse/SWIFT-892) so we can talk as a team about better approaches to this.

Sources/BSON/BSONDocument.swift
```diff
@@ -339,7 +303,53 @@ extension BSONDocument: BSONValue {
...
         buffer.writeBuffer(&doc)
     }
 }
+
+internal struct BSONDocumentStorage {
+    internal var buffer: ByteBuffer
+    internal init(_ buffer: ByteBuffer) { self.buffer = buffer }
+
+    /// Appends element to underlying BSON bytes, returns the size of the element appended: type + key + value
+    @discardableResult internal static func append(key: String, value: BSON, to buffer: inout ByteBuffer) -> Int {
```
patrickfreed, line 316: this can drop the `to` argument and no longer be static, since now the storage owns the buffer.

Sources/BSON/BSONDocument.swift
```diff
@@ -339,7 +303,53 @@ extension BSONDocument: BSONValue {
     }
 
     internal func write(to buffer: inout ByteBuffer) {
-        var doc = ByteBuffer(self._buffer.readableBytesView)
+        var doc = ByteBuffer(self.storage.buffer.readableBytesView)
         buffer.writeBuffer(&doc)
     }
 }
+
+internal struct BSONDocumentStorage {
```
patrickfreed, line 311: this isn't used outside of this file, so it can just be `private`.

Also, it could use a docstring describing its purpose. something like "A wrapper around a `ByteBuffer` providing various `BSONDocument`-specific utilities."

Sources/BSON/BSONDocument.swift
```diff
@@ -14,60 +14,68 @@ public struct BSONDocument {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BSON)
 
-    private var _buffer: ByteBuffer
+    internal var storage: BSONDocumentStorage
```
patrickfreed, line 17: this can remain private I think

PR #12, opened 2020-06-15
SWIFT-839 Evergreen project setup

kmahar:
This could probably be cleaned up more but I mainly wanted to get something up and working that we can improve upon later on.
I also set up evergreen PR testing, if I did it correctly we should see it show up when I open this...


.evergreen/config.yml
```diff
@@ -60,14 +61,50 @@ functions:
...
+  "lint":
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          SWIFT_VERSION=${SWIFT_VERSION} \
+            bash ${PROJECT_DIRECTORY}/.evergreen/install-tools.sh swiftlint
+          LINUX_SOURCEKIT_LIB_PATH=${PROJECT_DIRECTORY}/opt/swiftenv/versions/${SWIFT_VERSION}/usr/lib \
```
kmahar, line 94: this env var is necessary to let swiftlint find sourcekit on linux

.evergreen/config.yml
```diff
@@ -114,14 +160,31 @@ axes:
...
 - matrix_name: "tests-all"
+  display_name: "${swift-version} ${os-fully-featured}"
   matrix_spec:
     os-fully-featured: "*"
     swift-version: "*"
-  display_name: "${swift-version} ${os-fully-featured}}"
   tasks:
     - name: "test"
+
+- matrix_name: "sourcery"
```
kmahar, line 175: sourcery has to be run on macOS so this is split up. kind of annoying so we might want to consider doing this on travis or with a github action or something but for now i left it here

.evergreen/install-tools.sh
```diff
@@ -0,0 +1,51 @@
...
+# variables
+PROJECT_DIRECTORY=${PROJECT_DIRECTORY:-$PWD}
+SWIFT_VERSION=${SWIFT_VERSION:-5.2.4}
+INSTALL_DIR="${PROJECT_DIRECTORY}/opt"
+
+export SWIFTENV_ROOT="${INSTALL_DIR}/swiftenv"
+export PATH="${SWIFTENV_ROOT}/bin:$PATH"
+
+# usage: build_from_gh [name] [url] [tag]
+build_from_gh () {
```
kmahar, line 16: we have to build swiftformat and swiftlint ourselves on linux, but sourcery we can just download a binary since it runs on macOS

.swiftformat
```diff
@@ -1,4 +1,4 @@
---exclude .build/*
+--exclude .build/*,opt/*
```
kmahar, line 1: ignores the path we install stuff to on evg

.swiftlint.yml
```diff
@@ -44,6 +44,8 @@ opt_in_rules:
 excluded:
   - .build
   - Package.swift
+  - Tests/LinuxMain.swift
```
kmahar, line 47: this fails due to trailing commas

.evergreen/config.yml
```diff
@@ -60,14 +61,50 @@ functions:
...
+  - command: shell.exec
+    type: test
+    params:
+      working_dir: "src"
+      script: |
+        ${PREPARE_SHELL}
+        SWIFT_VERSION=${SWIFT_VERSION} \
+          bash ${PROJECT_DIRECTORY}/.evergreen/install-tools.sh swiftformat
+        SWIFT_VERSION=${SWIFT_VERSION} \
+        ${PROJECT_DIRECTORY}/opt/swiftformat/.build/release/swiftformat --lint --quiet .
```
patrickfreed, line 83: when the lint fails, `--quiet` makes it so that it just prints that the lint failed, not where it failed / why. I actually think we want `--verbose` for swiftformat to include this info. For swiftlint, `--quiet` seems to still print the errors so that should be fine.

kmahar: good catch. updated

PR #13, opened 2020-06-15
SWIFT-846: Add support for RegularExpression, Symbol, Timestamp, and Binary

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-846

Types types! a lot of the code should look familiar to the keen eyed swift driver developer but nonetheless adapted to use our new BSONValue protocol.


Sources/BSON/BSONBinary.swift
```diff
@@ -0,0 +1,206 @@
...
+        ])
+
+        try self.init(data: uuidData, subtype: BSONBinary.Subtype.uuid)
+    }
+
+    /// Initializes a `BSONBinary` instance from a `Data` object and a `UInt8` subtype.
+    /// - Throws:
+    ///   - `BSONError.InvalidArgumentError` if the provided data is incompatible with the specified subtype.
+    public init(data: Data, subtype: Subtype) throws {
+        if [Subtype.uuid, Subtype.uuidDeprecated].contains(subtype) && data.count != 16 {
```
patrickfreed, line 82: rather than duplicating this logic in each of the initializers, the ones for `Data` and `[UInt8]` should just construct `ByteBuffer` from them and vall the `ByteBuffer` initializer.

Sources/BSON/BSONBinary.swift
```diff
@@ -0,0 +1,206 @@
...
+            uuidt.0, uuidt.1, uuidt.2, uuidt.3,
+            uuidt.4, uuidt.5, uuidt.6, uuidt.7,
+            uuidt.8, uuidt.9, uuidt.10, uuidt.11,
+            uuidt.12, uuidt.13, uuidt.14, uuidt.15
+        ])
+
+        try self.init(data: uuidData, subtype: BSONBinary.Subtype.uuid)
+    }
+
+    /// Initializes a `BSONBinary` instance from a `Data` object and a `UInt8` subtype.
```
patrickfreed, line 78: this docstring should say `Subtype` instead of `UInt8`. Perhaps it should call out that this always creates a copy of the data regardless of CoW behavior.

PR #14, opened 2020-06-16
SWIFT-847: Code, OID, Null, Undef, MinMaxKey DBPtr

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-847

We don't have a ticket for the empty types null, min/max key, and undef so I included them here, they're all in BSONNull.swift since they're identical, bikeshed that file name? 
I had to add in a basic ObjectID to get DBPointer implemented, the ticket to create the ObjectID generator will flesh this out.


Sources/BSON/BSONCode.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    internal var bson: BSON { .codeWithScope(self) }
+
+    internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
+        let reader = buffer.readerIndex
+        guard let size = buffer.readInteger(endianness: .little, as: Int32.self) else {
+            throw BSONError.InternalError(message: "Cannot code with scope size")
+        }
+        guard size >= 14, (size - 4) < buffer.readableBytes else {
+            throw BSONError.InternalError(message: "Code with scope has size: \(size) but there "
```
patrickfreed, line 58: this error message only applies to the latter condition in this `guard`, so it needs to be split up so each condition can have an appropriate error message.

also, including a comment indicating how the 14 number was calculated would help future readers.

Sources/BSON/BSONCode.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+        guard let code = try String.read(from: &buffer).stringValue else {
+            throw BSONError.InternalError(message: "Cannot read code")
+        }
+        guard let scope = try BSONDocument.read(from: &buffer).documentValue else {
+            throw BSONError.InternalError(message: "Cannot read scope document")
+        }
+        guard (buffer.readerIndex - reader) == size else {
+            throw BSONError.InternalError(
+                message: "Stated size: \(size) is not correct read \(buffer.readerIndex - reader)"
```
patrickfreed, line 69: maybe replace "read" with ", actual size: " to make this clearer

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -124,6 +109,9 @@ final class BSONCorpusTests: BSONTestCase {
                         XCTFail("Unable to interpret canonical_bson as Data")
                         return
                     }
+                    let description = "\(testFile.description)-\(test.description)"
+
+                    print("VALID: \(description)")
```
patrickfreed, line 114: debug print here and below needs to be removed

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -215,10 +212,6 @@ final class BSONCorpusTests: BSONTestCase {
 
             if let decodeErrors = testFile.decodeErrors {
                 for test in decodeErrors where shouldRun(testFile.description, test.description) {
-                    let description = "\(testFile.description)-\(test.description)"
```
patrickfreed, line 230: I think this description still needs to be defined for the `XCTFail` below to compile

nbbeeken: Oops! good catch turning the git hook back on 😅

PR #15, opened 2020-06-16
SWIFT-843: BSONDocument+Collection+Sequence

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-843

Collection and Sequence conformance, mostly lifted from the driver with modifications for our new underlying ByteBuffer.


Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -28,6 +13,53 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+            _ = mutBuffer.readBytes(length: 1) // type indicator
+            _ = try mutBuffer.readCString() // BSON Key
+            guard let type = BSON.allBSONTypes[typeIndicator] else {
+                throw BSONError.InternalError(message: "Unable to read type: \(typeIndicator)")
+            }
+            return try type.read(from: &mutBuffer)
+        } catch {
+            fatalError("Cannot obtain current value: \(error.localizedDescription)")
+        }
+    }
```
nbbeeken, line 61: This presented a bit of an issue, since we only have read and not a "get" the ByteBuffer needs to be mutable in order to read out the current type so I created a copy that's mutable, works fine but I wasn't sure if that's ideal.

Sources/BSON/BSONDocument+Sequence.swift
```diff
@@ -0,0 +1,223 @@
...
+    public func dropFirst(_ n: Int) -> BSONDocument {
+        switch n {
+        case ..<0:
+            fatalError("Can't drop a negative number of elements from a document")
+        case 0:
+            return self
+        default:
+            // get all the key-value pairs from nth index on. subsequence will handle the case where n >= length of doc
+            // by creating an iter and calling advance until the end is reached. this is exactly what calling self.count
+            // would do in that situation via bson_count_keys, so no point in special casing self.count <= n here.
```
patrickfreed, line 61: this comment seems out of date and needs revising

Sources/BSON/BSONDocumentIterator.swift
patrickfreed: This property seems to be a relic of having to interact with libbson. If you look at all the existing usages, its always preceded by a call to `advance`. We can make this a single operation (i.e. `next`) to avoid the duplication, and just remove this property altogether I think.

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -108,6 +140,52 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+        guard endIndex >= startIndex else {
+            fatalError("endIndex must be >= startIndex")
+        }
+
+        var iter = BSONDocumentIterator(over: doc)
+
+        var excludedKeys: [String] = []
+
+        for _ in 0..<startIndex {
+            if let next = iter.next() {
```
patrickfreed, line 161: this would be a little more clear as a `guard`

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -108,6 +140,52 @@ public struct BSONDocumentIterator: IteratorProtocol {
             }
         }
     }
+
+    // uses an iterator to copy (key, value) pairs of the provided document from range [startIndex, endIndex) into a new
+    // document. starts at the startIndex-th pair and ends at the end of the document or the (endIndex-1)th index,
+    // whichever comes first.
+    internal static func subsequence(
```
patrickfreed, line 147: optimizing for performance is a non-goal of this project, so I think the strategy for implementing this is fine as of now. That said, we'll want to improve this in the future to just slice up the requisite bytes into a new `BSONDocument`, rather than doing 2 iterations over the document. I filed [SWIFT-911](https://jira.mongodb.org/browse/SWIFT-911) for the work covering that.

In the future, if we amend `BSONDocument` to contain further metadata about byte ranges, we could implement this without any walks over the document at all.

patrickfreed, line 147: This static method looks like it would be better as a regular method, though it seems to have been written like this in the original BSON library for a while. Do you recall the reasoning for that @kmahar ?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -109,6 +109,9 @@ final class BSONCorpusTests: BSONTestCase {
                         XCTFail("Unable to interpret canonical_bson as Data")
                         return
                     }
+                    let description = "\(testFile.description)-\(test.description)"
```
patrickfreed, line 112: debug

Sources/BSON/BSONDocument+Collection.swift
```diff
@@ -0,0 +1,67 @@
...
+        var iter = BSONDocumentIterator(over: self)
+
+        guard position != 0 else {
+            guard let (k, v) = try? iter.nextThrowing() else {
+                fatalError("Failed get current key and value at \(position)")
+            }
+            return (k, v)
+        }
+
+        for pos in 0...(position - 1) {
```
patrickfreed, line 50: this would be clearer as a `..<position`. Also, once that change is made, the above check for `position != 0` can go away.

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -108,6 +97,52 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+        of doc: BSONDocument,
+        startIndex: Int = 0,
+        endIndex: Int = Int.max
+    ) -> BSONDocument {
+        // TODO(SWIFT-911): Improve performance
```
patrickfreed, line 109: the format of our TODOs is typically `TODO: SWIFT-XXX <reason>`. We should probably stick to this for easy grepping in the future.

Sources/BSON/BSONDocumentIterator.swift
```diff
@@ -108,6 +97,52 @@ public struct BSONDocumentIterator: IteratorProtocol {
...
+                // we ran out of values
+                break
+            }
+            excludedKeys.append(next.key)
+        }
+
+        // skip the values between startIndex and endIndex. this has better performance than calling next, because
+        // it doesn't pull the unneeded key/values out of the iterator
+        for _ in startIndex..<endIndex {
+            if (try? iter.nextThrowing()) == nil {
```
patrickfreed, line 129: This, and really any conditional whose body unconditionally leads to a scope exit, should be written as a `guard`. Doing so allows for easy detection of invariants by just looking for the keyword, and because the compiler enforces that a `guard` must exit scope in its body, it decreases the complexity of the following guarded section. It also of course eliminates the possibility of forgetting a `break` / `return` / `throw`.

PR #16, opened 2020-06-17
SWIFT-844: ObjectID and Generator

nbbeeken:
https://jira.mongodb.org/browse/SWIFT-844

Fleshed out the remaining implementation of ObjectID

### DIFF
--- a/Sources/BSON/BSONObjectID.swift
+++ b/Sources/BSON/BSONObjectID.swift
@@ -1,23 +1,32 @@
+import Foundation
 import NIO
+import NIOConcurrencyHelpers
 
 /// A struct to represent the BSON ObjectID type.
 public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
+    internal static let LENGTH = 12
+
     /// This `BSONObjectID`'s data represented as a `String`.
-    public var hex: String { fatalError("ah") }
+    public var hex: String { self.oid.reduce("") { $0 + String(format: "%02x", $1) } }
 
-    public var description: String {
-        self.hex
-    }
+    public var description: String { self.hex }
 
+    /// ObjectID Bytes
     internal let oid: [UInt8]
+
+    internal static let generator = ObjectIDGenerator()
 
     /// Initializes a new `BSONObjectID`.
     public init() {
-        fatalError("ah")
+        self.oid = Self.generator.generate()
     }
 
     /// Initializes a new `BSONObjectID`.
     internal init(_ bytes: [UInt8]) {
+        precondition(
+            bytes.count == BSONObjectID.LENGTH,
+            "BSONObjectIDs must be \(BSONObjectID.LENGTH) bytes long, got \(bytes.count)"
+        )
         self.oid = bytes
     }
 
@@ -26,7 +35,22 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
     ///   - `BSONError.InvalidArgumentError` if string passed is not a valid BSONObjectID
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(_ hex: String) throws {
-        fatalError("ah")
+        guard hex.utf8.count == (BSONObjectID.LENGTH * 2) else {
+            throw BSONError.InvalidArgumentError(
+                message: "Cannot create ObjectId from \(hex). Length must be \(BSONObjectID.LENGTH * 2)"
+            )
+        }
+        var data = [UInt8](repeating: 0, count: 12)
+        for i in 0..<BSONObjectID.LENGTH {
+            let j = hex.index(hex.startIndex, offsetBy: i * 2)
+            let k = hex.index(j, offsetBy: 2)
+            let bytes = hex[j..<k]
+            guard let num = UInt8(bytes, radix: 16) else {
+                throw BSONError.InvalidArgumentError(message: "Invalid hexadecimal character \(bytes)")
+            }
+            data[i] = num
+        }
+        self = BSONObjectID(data)
     }
 }
 
@@ -36,13 +60,48 @@ extension BSONObjectID: BSONValue {
     internal var bson: BSON { .objectID(self) }
 
     internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
-        guard let bytes = buffer.readBytes(length: 12) else {
-            throw BSONError.InternalError(message: "Cannot read 12 bytes for BSONObjectID")
+        guard let bytes = buffer.readBytes(length: BSONObjectID.LENGTH) else {
+            throw BSONError.InternalError(message: "Cannot read \(BSONObjectID.LENGTH) bytes for BSONObjectID")
         }
         return .objectID(BSONObjectID(bytes))
     }
 
     internal func write(to buffer: inout ByteBuffer) {
         buffer.writeBytes(self.oid)
+    }
+}
+
+/// A class responsible for generating ObjectIDs for a given instance of this library
+/// An ObjectID consists of a random number for this process, a timestamp, and a counter
+internal class ObjectIDGenerator {
+    /// Random value is 5 bytes of the ObjectID
+    private let randomNumber: [UInt8]
+
+    /// Increment counter is only 3 bytes of the ObjectID
+    internal var counter: NIOAtomic<UInt32>
+
+    private static let COUNTER_MAX: UInt32 = 0xFFFFFF
+    private static let RANDOM_MAX: UInt64 = 0xFF_FFFF_FFFF
+
+    internal init() {
+        // 5 bytes of a random number
+        self.randomNumber = [UInt8](withUnsafeBytes(
+            of: UInt64.random(in: 0...ObjectIDGenerator.RANDOM_MAX), [UInt8].init
+        )[0..<5])
+        // 3 byte counter started randomly per process
+        self.counter = NIOAtomic<UInt32>.makeAtomic(value: UInt32.random(in: 0...ObjectIDGenerator.COUNTER_MAX))
+    }
+
+    internal func generate() -> [UInt8] {
+        // roll over counter
+        _ = self.counter.compareAndExchange(expected: ObjectIDGenerator.COUNTER_MAX + 1, desired: 0x00)
+        // fetch current timestamp
+        let timestamp = UInt32(Date().timeIntervalSince1970)
+        var buffer = [UInt8]()
+        buffer.reserveCapacity(BSONObjectID.LENGTH)
+        buffer += withUnsafeBytes(of: timestamp.bigEndian, [UInt8].init)
+        buffer += self.randomNumber
+        buffer += withUnsafeBytes(of: self.counter.add(1).bigEndian, [UInt8].init)[1..<4] // bottom 3 bytes
+        return buffer
     }
 }
--- a/Tests/BSONTests/BSONCorpusTests.swift
+++ b/Tests/BSONTests/BSONCorpusTests.swift
@@ -109,7 +109,6 @@ final class BSONCorpusTests: BSONTestCase {
                         XCTFail("Unable to interpret canonical_bson as Data")
                         return
                     }
-
                     // guard let cEJData = test.canonicalExtJSON.data(using: .utf8) else {
                     //     XCTFail("Unable to interpret canonical_extjson as Data")
                     //     return
@@ -132,7 +131,7 @@ final class BSONCorpusTests: BSONTestCase {
                     // TODO(SWIFT-867): Enable these lines when you can do subscript assignment
                     let nativeFromDoc = docFromCB.toArray()
                     let docFromNative = BSONDocument(fromArray: nativeFromDoc)
-                    expect(docFromNative.byteString).to(equal(cBData.byteString))
+                    expect(docFromNative.toByteString()).to(equal(cBData.toByteString()))
 
                     // native_to_canonical_extended_json( bson_to_native(cB) ) = cEJ
                     // expect(docFromCB.canonicalExtendedJSON).to(cleanEqual(test.canonicalExtJSON))
--- a/Tests/BSONTests/BSONObjectIDTests.swift
+++ b/Tests/BSONTests/BSONObjectIDTests.swift
@@ -1,0 +1,75 @@
+@testable import BSON
+import Foundation
+import Nimble
+
+extension BSONObjectID {
+    // timestamp
+    internal var timestamp: Int {
+        var value = Int()
+        _ = withUnsafeMutableBytes(of: &value) { self.oid[0..<4].reversed().copyBytes(to: $0) }
+        return value
+    }
+
+    // random value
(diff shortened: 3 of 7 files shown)


Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+internal class ObjectIDGenerator {
+    private let randomNumber: [UInt8]
+    private var counter: NIOAtomic<UInt32>
+    internal init() {
+        // 5 bytes of a random number
+        self.randomNumber = Array(withUnsafeBytes(
+            of: UInt64.random(in: 0...0xFF_FFFF_FFFF).bigEndian, [UInt8].init
```
patrickfreed, line 72: here and elsewhere, it's probably a bit clearer to just use the `.max` constants (e.g. `UInt64.max` in this case)

Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
...
+        buffer.writeBytes(self.oid)
+    }
+}
+
+internal class ObjectIDGenerator {
+    private let randomNumber: [UInt8]
+    private var counter: NIOAtomic<UInt32>
+    internal init() {
+        // 5 bytes of a random number
+        self.randomNumber = Array(withUnsafeBytes(
```
patrickfreed, line 71: is the second `Array` wrapping necessary? I think `withUnsafeBytes(..., [UInt8].init)` should be sufficient.

Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
...
+        // 3 byte counter started randomly per process
+        self.counter = NIOAtomic<UInt32>.makeAtomic(value: UInt32.random(in: 0..<0xFFFFFF))
+    }
+
+    internal func generate() -> [UInt8] {
+        // roll over counter
+        _ = self.counter.compareAndExchange(expected: 0xFFFFFF, desired: 0x00)
+        // fetch current timestamp
+        let timestamp = UInt32(Date().timeIntervalSince1970)
+        var buffer = [UInt8]()
```
patrickfreed, line 83: this array can also be initialized with 12 values via the `init(repeating:length:)`.

Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return .objectID(BSONObjectID(bytes))
+    }
+
+    internal func write(to buffer: inout ByteBuffer) {
+        buffer.writeBytes(self.oid)
+    }
+}
+
+internal class ObjectIDGenerator {
```
patrickfreed, line 66: this type could use a docstring explaining why its here and what it does.

Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
+import NIO
+import NIOConcurrencyHelpers
+
+/// A struct to represent the BSON ObjectID type.
+public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
+    /// This `BSONObjectID`'s data represented as a `String`.
+    public var hex: String { self.oid.reduce("") { $0 + String(format: "%02x", $1) } }
```
patrickfreed, line 8: some tests verifying that converting an `ObjectID` to/from a hex string happens losslessly would be nice.

Sources/BSON/BSONObjectID.swift
```diff
@@ -0,0 +1,89 @@
...
+    internal init(_ bytes: [UInt8]) {
+        self.oid = bytes
+    }
+
+    /// Initializes an `BSONObjectID` from the provided hex `String`.
+    /// - Throws:
+    ///   - `BSONError.InvalidArgumentError` if string passed is not a valid BSONObjectID
+    /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
+    public init(_ hex: String) throws {
+        guard hex.utf8.count == 24 else {
```
patrickfreed, line 30: instead of a magic number, we should define a static constant under `ObjectID` that is it's length (i.e. 12). Then we can do 2 * that constant here, and the local variable `len` can just go away.

Sources/BSON/BSONObjectID.swift
nbbeeken: I think that would reduce the randomness, not sure though. It's supposed to be a 5 byte number and if I go beyond 5 bytes for the upper limit than that upper limit has the chance to occur many times as opposed to just once, does that make sense?

Sources/BSON/BSONObjectID.swift
nbbeeken: I take a slice of it so it's an ArraySlice if I don't cast it, I can make it better using a [UInt8]() constructor

Sources/BSON/BSONObjectID.swift
nbbeeken: I'm extending the array below I can do this but how do I assign to array slices? or do I just loop and assign per index?

Sources/BSON/BSONObjectID.swift
patrickfreed: Oh I see. Honestly I just saw a bunch of F's and assumed you were typing the literal for the max value, but I see you were only going for 5 bytes instead of the full 8.

Sources/BSON/BSONObjectID.swift
patrickfreed: Oh good point. In this case it would probably be better to use [`reserveCapacity()`](https://developer.apple.com/documentation/swift/array/1538966-reservecapacity) here instead then.

Sources/BSON/BSONObjectID.swift
```diff
@@ -26,7 +29,22 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
     ///   - `BSONError.InvalidArgumentError` if string passed is not a valid BSONObjectID
     /// - SeeAlso: https://github.com/mongodb/specifications/blob/master/source/objectid.rst
     public init(_ hex: String) throws {
-        fatalError("ah")
+        guard hex.utf8.count == (BSONObjectID.LENGTH * 2) else {
+            throw BSONError.InvalidArgumentError(message: "Cannot create ObjectId from \(hex)")
```
kmahar, line 33: since we know the issue here (invalid length) maybe we could say that in the error message?

Sources/BSON/BSONObjectID.swift
```diff
@@ -26,7 +29,22 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
...
+        }
+        let len = hex.utf8.count / 2
+        var data = [UInt8](repeating: 0, count: 12)
+        for i in 0..<len {
+            let j = hex.index(hex.startIndex, offsetBy: i * 2)
+            let k = hex.index(j, offsetBy: 2)
+            let bytes = hex[j..<k]
+            if let num = UInt8(bytes, radix: 16) {
```
kmahar, line 41: nit: seems like a natural place to use `guard let num...` rather than if/else

Sources/BSON/BSONObjectID.swift
```diff
@@ -26,7 +29,22 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
...
+                data[i] = num
+            } else {
+                throw BSONError.InvalidArgumentError(message: "Cannot create UInt8 from \(bytes)")
+            }
+        }
+        self = BSONObjectID([UInt8](data))
```
kmahar, line 47: 1. isn't it already a `[UInt8]`? is the cast necessary?
2. I can't comment on the initializer that accepts bytes since it's old code but right now we could accidentally circumvent the length requirement internally by using that. maybe consider moving the length check/error into there, or adding a precondition in that init (evaluated in debug only) that makes sure we have 12 bytes

Sources/BSON/BSONObjectID.swift
```diff
@@ -36,7 +54,7 @@ extension BSONObjectID: BSONValue {
     internal var bson: BSON { .objectID(self) }
 
     internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
-        guard let bytes = buffer.readBytes(length: 12) else {
+        guard let bytes = buffer.readBytes(length: BSONObjectID.LENGTH) else {
             throw BSONError.InternalError(message: "Cannot read 12 bytes for BSONObjectID")
```
kmahar, line 58: should we use the length constant in the error message too?

Sources/BSON/ByteBuffer+BSON.swift
```diff
@@ -10,14 +10,27 @@ extension ByteBuffer {
...
+    internal func getCString(at offset: Int) throws -> String {
+        let key = try getBSONKey(at: offset).dropLast()
+        guard let string = String(bytes: key, encoding: .utf8) else {
+            throw BSONError.InternalError(message: "Failed to decode BSONKey as UTF8: \(key)")
+        }
+        return string
+    }
+
+    /// Returns the C String key including the null byte, for ease of comparison and byte counting
+    internal func getBSONKey(at offset: Int) throws -> [UInt8] {
```
kmahar, line 28: I think this name is confusing because C strings aren't just used for BSON keys. can we come up with a name that better clarifies what exactly this method does? for example, "getCStringBytes" ? I also don't see this method used outside of `getCString`, where will we be using it otherwise?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -109,6 +109,9 @@ final class BSONCorpusTests: BSONTestCase {
                         XCTFail("Unable to interpret canonical_bson as Data")
                         return
                     }
+                    let description = "\(testFile.description)-\(test.description)"
```
kmahar, line 112: this looks like debugging changes?

Tests/DisabledTests/BSONValueTests.swift
```diff
@@ -1,11 +1,9 @@
-import CLibMongoC
+@testable import BSON
```
kmahar, line 1: given that this file is still in "disabled", does it make sense to include these changes now or no?

Sources/BSON/BSONObjectID.swift
```diff
@@ -26,7 +29,22 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
...
+        let len = hex.utf8.count / 2
+        var data = [UInt8](repeating: 0, count: 12)
+        for i in 0..<len {
+            let j = hex.index(hex.startIndex, offsetBy: i * 2)
+            let k = hex.index(j, offsetBy: 2)
+            let bytes = hex[j..<k]
+            if let num = UInt8(bytes, radix: 16) {
+                data[i] = num
+            } else {
+                throw BSONError.InvalidArgumentError(message: "Cannot create UInt8 from \(bytes)")
```
kmahar, line 44: i think it would be a bit clearer to say something like "Invalid hexadecimal character x" here

Sources/BSON/BSONObjectID.swift
```diff
@@ -46,3 +69,29 @@ extension BSONObjectID: BSONValue {
         buffer.writeBytes(self.oid)
     }
 }
+
+/// A class responsible for generating ObjectIDs for a given instance of this library
+/// An ObjectID consists of a random number for this process, a timestamp, and a counter
+internal class ObjectIDGenerator {
+    private let randomNumber: [UInt8]
+    private var counter: NIOAtomic<UInt32>
```
kmahar, line 77: it might be useful to note somewhere, either here or below where it's used that we only use the three lower bytes of this value

Tests/BSONTests/BSONTests.swift
```diff
@@ -71,26 +71,42 @@ func toByteString(_ bytes: [UInt8]?) -> String {
...
 }
 
 public extension Array where Element == UInt8 {
-    var byteString: String {
-        BSONTests.toByteString(self)
+    func toByteString(ascii: Bool = true) -> String {
+        BSONTests.toByteString(self, ascii: ascii)
+    }
+
+    init?(hexString: String) {
```
kmahar, line 84: this looks very similar to our logic for creating an ObjectID from a hex string. Perhaps we need some internal helper method that we can use in both places?

Tests/BSONTests/BSONTests.swift
```diff
@@ -53,14 +53,14 @@ public func retrieveSpecTestFiles<T: Decodable>(
         }
 }
 
-func toByteString(_ bytes: [UInt8]?) -> String {
+func toByteString(_ bytes: [UInt8]?, ascii: Bool = false) -> String {
```
kmahar, line 56: document what the ascii option does?

Sources/BSON/BSON.swift
```diff
@@ -211,6 +211,67 @@ extension BSON {
     }
 }
 
+extension BSON {
```
patrickfreed, line 214: ditto this doesn't seem related to the ticket at hand here, are these needed for the `ObjectID` tests or something?

Tests/BSONTests/BSONObjectIDTests.swift
```diff
@@ -0,0 +1,81 @@
+@testable import BSON
+import Foundation
+import Nimble
+
+extension BSONObjectID {
+    // timestamp
+    internal var timestamp: Int {
```
patrickfreed, line 7: instead of all this bit shifting, I think it would be clearer to just use a temporrary `ByteBuffer` or `withUnsafeBytes` + `load` approach that we used somewhere else.

kmahar:
Just did another pass. I agree with all of Patrick's outstanding comments; I don't have any further comments myself right now though.

Sources/BSON/BSON.swift
nbbeeken: Yea to both comments, this was the last ticket that was carrying around 50+ commits of messy history I learned my rebase lesson on this project
Anyway cleaned it up, this just has OID changes now

Sources/BSON/BSONObjectID.swift
```diff
@@ -1,23 +1,32 @@
...
+
     /// Initializes a new `BSONObjectID`.
     public init() {
-        fatalError("ah")
+        self.oid = Self.generator.generate()
     }
 
     /// Initializes a new `BSONObjectID`.
     internal init(_ bytes: [UInt8]) {
+        precondition(
```
patrickfreed, line 26: this precondition will only be triggered in debug mode, which is why we generally use `fatalError`. Is that intentional here?

kmahar: that was a suggestion by me - since this is an internal method and we only call it from places that already validate the length beforehand, it seems like overkill to check twice in release mode. but at the same time it seemed like this method should do *some* kind of validation

nbbeeken: It is intention in that @kmahar suggested it, I think because this is internal and used currently only when reading out of the ByteBuffer, which has already asserted it has read 12 bytes

patrickfreed: sgtm

PR #17, opened 2020-06-17
SWIFT-848: Update README

nbbeeken:
Fix up the readme, the logo is just for fun! I don't think we should keep it 😅


README.md
```diff
@@ -1,7 +1,100 @@
...
+
+## Installation
+
+The library supports use with Swift 5.1+. The minimum macOS version required to build the library is 10.14. The library is tested in continuous integration against macOS 10.14, Ubuntu 16.04, and Ubuntu 18.04.
+
+Installation is supported via [Swift Package Manager](https://swift.org/package-manager/).
+
+### Install BSON
+
+The driver contains two modules to support a variety of use cases: an asynchronous API in `MongoSwift`, and a synchronous API in `MongoSwiftSync`. The modules share a BSON implementation and a number of core types such as options `struct`s.
```
kmahar, line 38: this sentence needs to be updated

README.md
```diff
@@ -1,7 +1,100 @@
...
+
+To install the driver, add the package and relevant module as a dependency in your project's `Package.swift` file:
```
kmahar, line 40: this sentence should also be updated, maybe just "driver" -> "library" and drop the "relevant" since there is only one module

README.md
```diff
@@ -1,7 +1,100 @@
...
+// swift-tools-version:5.1
+import PackageDescription
+
+let package = Package(
+    name: "MyPackage",
+    dependencies: [
+        .package(url: "https://github.com/mongodb/swift-bson.git", from: "VERSION.STRING.HERE"),
+    ],
+    targets: [
+        // Async module
```
kmahar, line 52: this comment should be deleted

README.md
```diff
@@ -1,7 +1,100 @@
...
+}
+print(doubled) // prints `[2, 4, 6, 8]`
+```
+
+Note that `BSONDocument` conforms to `Collection`, so useful methods from [`Sequence`](https://developer.apple.com/documentation/swift/sequence) and [`Collection`](https://developer.apple.com/documentation/swift/collection) are all available. However, runtime guarantees are not yet met for many of these methods.
+
+## Development Instructions
+
+See our [development guide](https://github.com/mongodb/mongo-swift-driver/blob/master/Guides/Development.md) for the MongoDB driver to get started.
+To run the tests for the bson library you can make use of the Makefile and run: `make test-pretty`
```
kmahar, line 96: bson -> BSON

maybe worth also clarifying `make test` vs `make test-pretty` (latter only works on macOS I think)

nbbeeken:
> just some small nits but mostly LGTM, logo is cute. do the bits translate to an actual document?

Thanks! 😄  it's just the same 1's and 0's from bsonspec.org, spent a lotta time there ⛏️

PR #19, opened 2020-07-02
SWIFT-915 Port BSONEncoder and BSONDecoder

patrickfreed:
[SWIFT-915](https://jira.mongodb.org/browse/SWIFT-915)

This PR ports `BSONEncoder` and `BSONDecoder` from the existing BSON library to `swift-bson`. As part of that, it also conforms the various BSON types (including `BSON` itself) to `Codable`, though most of the conformances will only work with `BSONEncoder/BSONDecoder` as is the case in the existing BSON library.

Almost all the changes are copy/paste, with the exception of some tests that were commented out that relied on extJSON or decimal128.


nbbeeken:
I realize we don't actually have EVG do any macos tests, it fails to compile on Mac right now, you need to add platforms to the package file:
```
    platforms: [
        .macOS(.v10_14)
    ],
```

Sources/BSON/BSONBinary.swift
```diff
@@ -188,4 +188,12 @@ extension BSONBinary: BSONValue {
         }
         buffer.writeBytes(self.data.readableBytesView)
     }
+
+    public init(from decoder: Decoder) throws {
+        throw getDecodingError(type: Self.self, decoder: decoder)
+    }
+
+    public func encode(to: Encoder) throws {
```
kmahar, line 196: nit: can we name the encoder rather than just using the `to` label? e.g. `to encoder: Encoder`.  same elsewhere

also a question, since these conformances are all identical can we save ourselves duplication and put the erroring behavior as the default conformance on `BSONValue`?

Sources/BSON/BSONError.swift
```diff
@@ -67,3 +67,68 @@ internal func BSONIterationError(
     error += " \(message)"
     return BSONError.InternalError(message: error)
 }
+
+/// Error thrown when a BSONValue type introduced by the driver (e.g. BSONObjectID) is encoded not using BSONEncoder
+internal func bsonEncodingUnsupportedError<T: BSONValue>(value: T, at codingPath: [CodingKey]) -> EncodingError {
+    let description = "Encoding \(T.self) BSONValue type with a non-BSONEncoder is currently unsupported"
```
kmahar, line 73: should we just say "BSON" instead of "BSONValue", since BSONValue is internal? (same applies to methods below)

Sources/BSON/BSONObjectID.swift
```diff
@@ -69,6 +69,28 @@ extension BSONObjectID: BSONValue {
...
+
+    public init(from decoder: Decoder) throws {
+        // assumes that the BSONObjectID is stored as a valid hex string.
+        let container = try decoder.singleValueContainer()
+        let hex = try container.decode(String.self)
+        guard let oid = try? BSONObjectID(hex) else {
+            throw DecodingError.dataCorrupted(
+                DecodingError.Context(
+                    codingPath: decoder.codingPath,
+                    debugDescription: "Invalid ObjectID hex string. Got: \(hex)"
```
kmahar, line 81: might want to just use the error message string here now that we throw more descriptive messages

Sources/BSON/CodableNumber.swift
```diff
@@ -0,0 +1,193 @@
...
+        }
+        // we could consider trying a Decimal128 here. However,
+        // it's not clear how we could support decoding something
+        // stored as Decimal128 back to a UInt without access
+        // to libbson internals.
+        return nil
+    }
+}
+
+/// Override the default initializer due to a runtime assertion that fails
```
kmahar, line 138: I wonder if we still still need to do this? same for Float below. I think those comments are really old, probably pre Swift 5.

Sources/BSON/CodingStrategies.swift
```diff
@@ -0,0 +1,218 @@
+import Foundation
+
+/// Protocol indicating a set of options can be used to configure `BSONEncoder` and `BSONDecoder`.
+public protocol CodingStrategyProvider {
```
kmahar, line 4: I kind of wonder if this entire file still belongs in the driver itself, as this is all meant to support setting strategies for clients/dbs/collections. do you think it has any value for general BSON usage?

Sources/BSON/BSONDocument+Codable.swift
```diff
@@ -0,0 +1,31 @@
...
+    public func encode(to encoder: Encoder) throws {
+        guard let bsonEncoder = encoder as? _BSONEncoder else {
+            throw bsonEncodingUnsupportedError(value: self, at: encoder.codingPath)
+        }
+
+        // directly add the `BSONDocument` to the encoder's storage.
+        bsonEncoder.storage.containers.append(self)
+    }
+
+    /// This method will work with any `Decoder`, but for non-BSON
```
kmahar, line 14: can we update this comment? it seems quite out of date.

Sources/BSON/BSONBinary.swift
patrickfreed: Done and done. 

I had assumed the extension wasn't possible because we didn't do it already, but it seems to work just fine.

Sources/BSON/BSONDocument+Codable.swift
patrickfreed: this entire file can actually just be replaced by the default `Codable` conformance now

Sources/BSON/BSONError.swift
patrickfreed: done

Sources/BSON/BSONObjectID.swift
patrickfreed: done

Sources/BSON/CodableNumber.swift
patrickfreed: done. seems to work fine.

Sources/BSON/CodingStrategies.swift
patrickfreed: I think it's still valuable in BSON-only contexts, mostly just to ensure that the strategies are consistent between the encoders/decoders. With e.g. `BSONCoderOptions`, a user could define a single options struct that they pass to both their `BSONEncoder()` and `BSONDecoder()` calls that ensures they both speak the same format.

kmahar: sgtm

Sources/BSON/BSONDocument+Codable.swift
kmahar: does the special-casing we were doing here e.g.
```
// directly add the `BSONDocument` to the encoder's storage.
        bsonEncoder.storage.containers.append(self)
```
not actually matter?

patrickfreed: Yeah, in `BSONEncoder` we don't invoke the `encode` method if the type is something we have a strategy for (e.g. `Date`), `BSON`, `BSONValue`, or `[BSONValue]`.

patrickfreed: Hmm, I actually think the `Decodable` conformance is necessary though, since we don't have any special casing for `BSONValue`s in `BSONDecoder`.

patrickfreed: Oh actually, we [do have special casing](https://github.com/mongodb/mongo-swift-driver/blob/master/Sources/MongoSwift/BSON/BSONDecoder.swift#L497), so the conformance isn't necessary.

PR #20, opened 2020-07-08
SWIFT-893 Expose timestamp property on BSONObjectID

nspektor:


Tests/BSONTests/BSONObjectIDTests.swift
```diff
@@ -48,7 +41,7 @@ final class BSONObjectIDTests: BSONTestCase {
 
     func testFieldAccessors() throws {
         let oid = try BSONObjectID("FEEEEEEEFBBBBBBBBBFAAAAA")
-        expect(oid.timestamp).to(equal(0xFEEE_EEEE))
+        expect(oid.timestamp).to(equal(Date(timeIntervalSince1970: 0xFEEE_EEEE)))
```
nspektor, line 44: Wasn't sure what to do here, since FEEEEEEE isn't really meant to represent a meaningful date and this test isn't directly meant to test that the timestamp is set correctly. Should I just delete this line? Leave it the way I did it? or more closely mimic the way `testTimestampCreation` checks the timestamp and ignore seconds?

Sources/BSON/BSONObjectID.swift
```diff
@@ -11,6 +11,16 @@ public struct BSONObjectID: Equatable, Hashable, CustomStringConvertible {
...
     public var description: String { self.hex }
 
+    /// The timestamp portion of this `BSONObjectID` represented as a `Date`.
+    public var timestamp: Date {
+        var value = Int()
+        _ = withUnsafeMutableBytes(of: &value) {
+            self.oid[0..<4].reversed().copyBytes(to: $0)
+        }
+        return Date(timeIntervalSince1970: TimeInterval(value))
+        print(self.timestamp)
```
patrickfreed, line 21: debug print needs to be removed.

Tests/BSONTests/BSONObjectIDTests.swift
patrickfreed: I think these tests are probably sufficient.

Sources/BSON/BSONObjectID.swift
nspektor: oh yeah, thanks for catching that

Tests/BSONTests/BSONObjectIDTests.swift
```diff
@@ -62,8 +55,8 @@ final class BSONObjectIDTests: BSONTestCase {
     }
 
     func testTimestampCreation() throws {
```
patrickfreed, line 57: Running this test locally, it takes 0.001 seconds to run, so I think the risk of us rolling over to the next second should be pretty low. To ensure complete testing, lets add the seconds back in here. If it fails too often we can just revert it.

Tests/BSONTests/BSONObjectIDTests.swift
patrickfreed, line 44: Now that I think about it, a minor improvement to this would be to construct the `Date` from a format string that corresponds to the number of seconds, rather than just initializing from the hex itself. 

From just looking at the date now, we can use the following:
- hex substring for date: `5F07445C`
- format string for date: "2020-07-09 12:22:52"

You'll need to create a `DateFormatter` like is used below and use `format.date` to construct the `Date` used as the "expected" one.

patrickfreed: Looks like you may need to set the timezone to UTC to ensure this test works. From the "Working with fixed format date representations" section of the [`DateFormatter` docs](https://developer.apple.com/documentation/foundation/dateformatter), you can do that with `format.timeZone = TimeZone(secondsFromGMT: 0)`.

Tests/BSONTests/BSONObjectIDTests.swift
nspektor: done

Tests/BSONTests/BSONObjectIDTests.swift
nspektor: this should be fixed now

PR #21, opened 2020-07-13
SWIFT-916 Implement JSON Enum

nspektor:


Sources/BSON/JSON.swift
```diff
@@ -0,0 +1,138 @@
+import Foundation
+
+/// Enum representing a JSON value
```
patrickfreed, line 3: can we add ", used internally for modeling JSON during extJSON parsing/generation" to this docstring? To people new to the codebase, it might not be immediately obvious why we need a JSON enum in a BSON library.

Sources/BSON/JSON.swift
```diff
@@ -0,0 +1,138 @@
...
+internal enum JSON: Codable {
+    case number(Double)
+    case string(String)
+    case bool(Bool)
+    indirect case array([JSON])
+    indirect case object([String: JSON])
+    case null
+
+    /// Initialize a `JSON` from a decoder
```
patrickfreed, line 12: as a general style thing, we try to write the docstrings in full sentences that end in periods. We don't follow this everywhere unfortunately, but in all new code we try to keep to it.

Sources/BSON/JSON.swift
```diff
@@ -0,0 +1,138 @@
...
+    /// Try to decode into each of the JSON types one by one until one succeeds or
+    /// if throw an error indicating that the input is not a valid `JSON` type
```
patrickfreed, line 14: "if throw" -> "throws"

Sources/BSON/JSON.swift
```diff
@@ -0,0 +1,138 @@
...
+extension JSON: ExpressibleByDictionaryLiteral {
+    public init(dictionaryLiteral elements: (String, JSON)...) {
+        self = .object([String: JSON](uniqueKeysWithValues: elements))
+    }
+}
+
+/// Value Getters
+extension JSON {
+    /// If this `JSON` is a `.double`, return it as a `Double`. Otherwise, return nil.
+    public var doubleValue: Double? {
```
patrickfreed, line 98: since `JSON` is `internal` itself, these `public` ACLs really are `internal` too, so can we update them to match? Ditto for other ACLs in this file.

Sources/BSON/JSON.swift
nspektor: yep

Sources/BSON/JSON.swift
nspektor: fixed

Sources/BSON/JSON.swift
nspektor: done

Sources/BSON/JSON.swift
nspektor: fixed

Tests/BSONTests/JSONTests.swift
```diff
@@ -0,0 +1,72 @@
...
+    func testInteger() throws {
+        // Initializing a JSON with an int works, but it will be cast to a double.
+        let intJSON: JSON = 12
+        let encoded = try encoder.encode([intJSON])
+        /* JSONEncoder currently cannot encode non-object/array top level values.
+         To get around this, the generated JSON will need to be wrapped in an array
+         and unwrapped again at the end as a workaround.
+         This workaround can be removed when Swift 5.3 is the minimum supported version by the BSON library. */
+        expect(Double(String(data: encoded.dropFirst().dropLast(), encoding: .utf8)!)!)
+            .to(beCloseTo(12))
```
kmahar, line 20: ah! `beCloseTo` is a nice find! I needed something like this in a recent PR but didn't see this existed so I rounded and then asserted: 
https://github.com/mongodb/mongo-swift-driver/pull/508/files#diff-e91729b2f3ded39497c56b4c18163e40R348

I'll have to switch over to using this.

Sources/BSON/JSON.swift
```diff
@@ -0,0 +1,139 @@
...
+    case string(String)
+    case bool(Bool)
+    indirect case array([JSON])
+    indirect case object([String: JSON])
+    case null
+
+    /// Initialize a `JSON` from a decoder.
+    /// Tries to decode into each of the JSON types one by one until one succeeds or
+    /// throws an error indicating that the input is not a valid `JSON` type.
+    public init(from decoder: Decoder) throws {
```
patrickfreed, line 16: `init` and `encode` here can be made `internal` as well.

nspektor: oh okay, I misunderstood our slack conversation from earlier. fixed now

kmahar:
On a related note, I recently learned that if you ever remove an email address from your GitHub account, any commits you made under that address will stop being attributed to you.

nspektor:
So should I set my personal (permanent) email as my email in my local git config?

kmahar:
I think that seems like the easiest approach. (Though you can also solve it by never deleting your MongoDB email address from your account even once its deactivated post internship.)

PR #22, opened 2020-07-13
SWIFT-917 Add sketch of ExtendedJSONDecoder public API

nspektor:


Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+/// Decoder for ExtendedJSON
+public class ExtendedJSONDecoder {
+    /// Initialize a Decoder
+    public init() {
+        fatalError("unimplemented")
+    }
+
+    /// Takes in JSON as `Data` encoded with `.utf8` and runs it through a `JSONDecoder` to get an
```
kmahar, line 9: I think this is an implementation detail we don't need to expose to users in our public documentation. the existence of a `JSON` type should be invisible to them entirely, and they don't really need to know how we got it. 

instead, we should state what this method's purpose is from a user's perspective: something like "Decodes an instance of the requested type `T` from the provided extended JSON data."

that said, the comment is definitely useful to a developer working on this library, so perhaps you could move it into the top of the method body?

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+/// Decoder for ExtendedJSON
```
kmahar, line 2: let's make this a complete sentence

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
...
+public class ExtendedJSONDecoder {
+    /// Initialize a Decoder
```
kmahar, line 4: complete sentence needed here as well, and let's use the full type name?

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    /// Takes in JSON as `Data` encoded with `.utf8` and runs it through a `JSONDecoder` to get an
+    /// instance of the `JSON` enum. Then a `BSON` enum instance is created via the `JSON`.
+    /// The `BSON is then passed through a `BSONDecoder` where it is outputted as a `T`
+    ///
+    /// Data --> JSON --> BSON --> T
+    ///
+    /// - Parameters:
+    ///   - type: Codable type to decode the input into.
+    ///   - data: `Data` which represents the `JSON` that will be decoded.
```
kmahar, line 17: while `JSON` is actually a type, it's not one that the user knows about, so I think we can omit the backticks here and the next line. otherwise they might think there is some type they are supposed to be passing in here rather than a `Data`.

Sources/BSON/ExtendedJSONDecoder.swift
nspektor: okay! done :)

Sources/BSON/ExtendedJSONDecoder.swift
nspektor: done

Sources/BSON/ExtendedJSONDecoder.swift
nspektor: done

Sources/BSON/ExtendedJSONDecoder.swift
nspektor: done

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
...
+    public init() {
+        fatalError("unimplemented")
+    }
+
+    /// Decodes an instance of the requested type `T` from the provided extended JSON data.
+    ///
+    /// - Parameters:
+    ///   - type: Codable type to decode the input into.
+    ///   - data: `Data` which represents the JSON that will be decoded.
+    /// - Returns: Decoded representation of the JSON input as an instance of T.
```
kmahar, line 14: the T needs backticks

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
...
+    /// - Throws:
```
kmahar, line 15: whoops, I missed before that we don't actually have an error here. I think you could probably just copy the equivalent error info from `BSONDecoder`.

nspektor: Yeah I thought I would just fill this in once I actually do the implementation, but I can do this instead

Sources/BSON/ExtendedJSONDecoder.swift
nspektor: done

Sources/BSON/ExtendedJSONDecoder.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+/// `ExtendedJSONDecoder` facilitates the decoding of ExtendedJSON into `Decodable` values.
+public class ExtendedJSONDecoder {
+    /// Initialize an `ExtendedJSONDecoder`.
+    public init() {
+        fatalError("unimplemented")
+    }
+
+    /// Decodes an instance of the requested type `T` from the provided extended JSON data.
```
patrickfreed, line 9: I think a `- SeeAlso:` with the following [link](https://docs.mongodb.com/manual/reference/mongodb-extended-json/) would be useful here.

nspektor: added

PR #23, opened 2020-07-15
SWIFT-918 Extended JSON to BSON Conversion for Int32

nspektor:


Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,34 @@
...
+                return nil
+            }
+            self = int
+        case let .object(obj):
+            // canonical extended JSON
+            guard obj.count == 1 else {
+                throw BSONError.InternalError(message: "Not a valid Int32")
+            }
+            guard let str = obj["$numberInt"]?.stringValue else {
+                throw BSONError.InternalError(message: "Not a valid Int32")
```
patrickfreed, line 21: in this case, we just want to return `nil`, since it means that the given JSON doesn't correspond to an `Int32`.

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,34 @@
...
+        switch json {
+        case let .number(n):
+            // relaxed extended JSON
+            guard let int = Int32(exactly: n) else {
+                return nil
+            }
+            self = int
+        case let .object(obj):
+            // canonical extended JSON
+            guard obj.count == 1 else {
```
patrickfreed, line 17: this check should come after the check for `$numberInt`, since otherwise we may throw errors when working with extended json that doesnt correspond to `Int32`.

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,34 @@
...
+                throw BSONError.InternalError(message: "Not a valid Int32")
+            }
+            guard let str = obj["$numberInt"]?.stringValue else {
+                throw BSONError.InternalError(message: "Not a valid Int32")
+            }
+            guard let int = Int32(str) else {
+                throw BSONError.InternalError(message: "Not a valid Int32")
```
patrickfreed, line 24: These errors will be thrown by `ExtendedJSONDecoder`, so they need to be `DecodingError`. `BSONError.InternalError` is reserved basically for invariant failures or other unexplainable errors, usually indicating bugs in our code.

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -0,0 +1,27 @@
+@testable import BSON
+import Foundation
+import Nimble
+import NIO
+import XCTest
+
+open class ExtendedJSONConversionTestCase: XCTestCase {
```
patrickfreed, line 7: this should subclass from `BSONTestCase`

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        // Nil cases
+        expect(try Int32(fromExtJSON: JSON.number(Double(Int32.max) + 1))).to(beNil())
+        expect(try Int32(fromExtJSON: JSON.bool(true))).to(beNil())
+
+        // Error cases
+        expect(try Int32(fromExtJSON: ["$numberInt": "5", "extra": true]))
+            .to(throwError(errorType: BSONError.InternalError.self))
+        expect(try Int32(fromExtJSON: ["$numberInt": .number(Double(Int32.max) + 1)]))
+            .to(throwError(errorType: BSONError.InternalError.self))
+        expect(try Int32(fromExtJSON: ["bad": "5"]))
```
patrickfreed, line 22: this should be a nil case, since it just doesnt match up with the extended JSON format of an `Int32`. The other two are errors because they indicate that an `Int32` is being specified, but then provide values that are incompatible with the type (e.g. overflowing `Int32`) or specify an invalid format (an extra field).

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,44 @@
...
+        case let .object(obj):
+            // canonical extended JSON
+            guard let value = obj["$numberInt"]?.stringValue else {
+                return nil
+            }
+            guard obj.count == 1 else {
+                throw DecodingError.dataCorrupted(
+                    DecodingError.Context(
+                        codingPath: [],
+                        debugDescription: "Not a valid Int32"
```
patrickfreed, line 24: this description could be a little more specific to what is actually going on, e.g. "Expected only "$numberInt" key, found extra keys: \(obj.keys)"

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,44 @@
...
+                    )
+                )
+            }
+            guard let int = Int32(value) else {
+                throw DecodingError.dataCorrupted(
+                    DecodingError.Context(
+                        codingPath: [],
+                        debugDescription: "Not a valid Int32"
```
patrickfreed, line 32: Similarly, this could be a little more specific: "Could not parse a valid Int32 from \\"\\(value)\\""

Sources/BSON/Integers+BSONValue.swift
patrickfreed: Hmm, good point. I think `[]` as the `codingPath` makes sense, since this error is occurring during preprocessing. We probably will want to include the key we're at in the error message though, so I think this initializer should have its own `keyPath: [String]` that gets built up as you recursively descend into the BSON structure (more recent keys being appended to the path as we go down).

Then the error messages could include something like: "doc.subdoc.array.1: could not parse Int32 from "bladfljk"".

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -0,0 +1,26 @@
...
+        expect(try Int32(fromExtJSON: 5)).to(equal(5))
+        expect(try Int32(fromExtJSON: ["$numberInt": "5"])).to(equal(5))
+
+        // Nil cases
+        expect(try Int32(fromExtJSON: JSON.number(Double(Int32.max) + 1))).to(beNil())
+        expect(try Int32(fromExtJSON: JSON.bool(true))).to(beNil())
+        expect(try Int32(fromExtJSON: ["bad": "5"])).to(beNil())
+
+        // Error cases
+        expect(try Int32(fromExtJSON: ["$numberInt": "5", "extra": true]))
```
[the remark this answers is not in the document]
nspektor: Is that the behavior we want? would it make more sense to throw here?

Sources/BSON/Integers+BSONValue.swift
nspektor: keyPath would only get set through the extended json initializers on object and array right? 

Would we want it to be set for the cases like `["$numberInt":"bla"]` and have the error there be `"$numberInt: could not parse Int32 from "bla""`?

patrickfreed: >keyPath would only get set through the extended json initializers on object and array right?

Yep, exactly.

>Would we want it to be set for the cases like ["$numberInt":"bla"] and have the error there be "$numberInt: could not parse Int32 from "bla""?

Nope, those keys are really part of the "value" of the integer.

Tests/BSONTests/ExtendedJSONConversionTests.swift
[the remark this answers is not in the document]
patrickfreed: Yeah throwing for the `["$numberInt": 5]` case is what we want. Note that 5 is an integer, not a string [as is required by the extended JSON spec](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table). In my comment I meant that as-written I believe that input would return nil but that it should be throwing.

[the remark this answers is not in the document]
nspektor: okay I fixed it and added an error case

Sources/BSON/Integers+BSONValue.swift
nspektor: done

Sources/BSON/Integers+BSONValue.swift
nspektor: done

Sources/BSON/Integers+BSONValue.swift
nspektor: okay I added this and added a test that passes one in. should I try to test that the debug message is as expected? or is that unnecessary

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,66 @@
...
+    *   - `keyPath`: an array of `String`s containing the enclosing JSON keys of the current json being passed in.
+    *              This is used for error messages.
+    *
+    * Returns:
+    *   - `nil` if the provided value is not an `Int32`.
+    *
+    * Throws:
+    *   - `DecodingError` if `json` is a partial match or is malformed.
+    */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]? = nil) throws {
```
patrickfreed, line 18: I think we always want a keyPath provided, since this will only be used in the context of decoding things. So I think we want this to just be a `[String]` with no default value.

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,66 @@
...
+                )
+            }
+            guard let int = Int32(str) else {
+                let debugStart = keyPath != nil
+                        ? "\(keyPath!.joined(separator:".")): "  
+                        : ""
+                throw DecodingError.dataCorrupted(
+                    DecodingError.Context(
+                        codingPath: [],
+                        debugDescription: "\(debugStart): could not parse `Int32` from \"\(str)\"."
```
patrickfreed, line 54: We alway want this prefix appended to `DecodingError`s returned from this initializer, so that's why I was suggesting a helper function or extension of `DecodingError` to aid in that process. Then all three cases where we throw errors could just call through to that and automatically get the path appended properly.

Sources/BSON/Integers+BSONValue.swift
patrickfreed: Now that I think about this some more, we may not always have a key path, since we'll be supporting decoding non-array/object stuff at the top level. So I think you're right that this should be optional, though I still think we can remove this default value so that we don't forget to pass it through as necessary.

Sources/BSON/Integers+BSONValue.swift
patrickfreed: Similar to my comment above, we don't _always_ want to prefix the key path, but if it is present then we always want to. So the same suggestion applies but whatever helper would need to take in a `[String]?`.

Sources/BSON/Integers+BSONValue.swift
patrickfreed: I think that's unnecessary for now, but we can do that for the `Document` / `Array` cases.

Sources/BSON/Integers+BSONValue.swift
patrickfreed: I think I've changed my mind again and I promise this will be the last: I think we actually don't want it to be optional, but for top level stuff just have an empty array `[]`. In the error we can just omit the prefix if the array is empty.

nspektor: if we remove the `=nil` do we have to manually pass in `nil` each time?

Sources/BSON/Integers+BSONValue.swift
nspektor: yep, I think I addressed this in the most recent commit :)

Sources/BSON/BSONValue.swift
```diff
@@ -14,6 +14,11 @@ internal protocol BSONValue: Codable {
 
     /// Writes this value's BSON byte representation to the provided ByteBuffer.
     func write(to buffer: inout ByteBuffer)
+
+    /// Initializes a corresponding `BSON` from the provided extendedJSON.
+    /// Currently commented out to avoid "not implemented" errors while implementing the
+    /// initializer one by one on each BSONValue.
+    // internal init(fromExtJSON json: JSON) throws
```
nspektor, line 21: should I just delete this for now? and add it back in after all the implementations

Sources/BSON/BSONError.swift
```diff
@@ -42,6 +42,19 @@ public enum BSONError {
     }
 }
 
+/// Standardize the errors emitted by BSONValue initializers.
+internal func BSONDecoderError(
```
patrickfreed, line 46: this function name makes it seem like it's a different error type or struct, rather than just a helper function. Instead, let's add it as an extension to `DecoderError` like we have for the other ones. e.g. `DecoderError.extendedJSONError(...)`.

Sources/BSON/BSONError.swift
```diff
@@ -42,6 +42,19 @@ public enum BSONError {
...
+    codingPath: [CodingKey] = [],
+    keyPath: [String],
+    debugDescription: String
+) -> DecodingError {
+    let debugStart = keyPath.joined(separator: ".")
+    return DecodingError.dataCorrupted(DecodingError.Context(
+        codingPath: codingPath,
+        debugDescription: debugStart + debugDescription
```
patrickfreed, line 54: as written, this will produce strings like "x.yExpected only \"$numberInt\" key, found extra keys in object: []", there needs to be a colon and space here. For cases where the key path is empty, make sure to omit them though, otherwise we'll have ": <rest of error message>".

Sources/BSON/BSONError.swift
nspektor: I changed it. followed the _ format from the other extensions. is that fine?

Sources/BSON/BSONError.swift
nspektor: fixed

Sources/BSON/BSONValue.swift
patrickfreed: I think it's fine to just leave it in.

Sources/BSON/BSONError.swift
```diff
@@ -42,6 +42,22 @@ public enum BSONError {
     }
 }
 
+extension DecodingError {
+    /// Standardize the errors emitted by BSONValue initializers.
+    internal static func _extendedJSONError(
+        codingPath: [CodingKey] = [],
```
patrickfreed, line 48: This array will always be `[]`, so I think we can omit it.

To loop in @kmahar: we're currently using an empty `codingPath` in the errors thrown as part of the `JSON` -> `BSONvalue` conversion since that stage is technically part of "preprocessing" and the user hasn't requested any keys yet. We could alternatively manage our own `codingPath` exactly as we're doing right now with `keyPath`, but I wasn't sure if that was the accurate thing to do.

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,52 @@
...
+            self = int
+        case let .object(obj):
+            // canonical extended JSON
+            guard let value = obj["$numberInt"] else {
+                return nil
+            }
+            guard obj.count == 1 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$numberInt\" key, found extra keys in object: \(obj)"
```
patrickfreed, line 34: this will include the entire object, but we only really care about the extra keys. To keep the error message brief, we can just do `obj.keys` here.

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -1,6 +1,52 @@
 import NIO
 
 extension Int32: BSONValue {
+    /*
+     * Initializes an `Int32` from ExtendedJSON
```
kmahar, line 5: nit: add period at end of line

Sources/BSON/BSONError.swift
patrickfreed: Yeah, just instead of including it in the error message we'd have it as coding path. We originally opted not to do this because the `codingPath` is typically meant to represent the path that the the user took as part of decoding through keyed containers. The `keyPath` here is just where we are in the document when encountering an error during preprocessing.

kmahar: Ok, yeah I agree with not using the coding path for that, leaving it in the error message makes sense.

nspektor: okay sounds good, I removed it

Sources/BSON/Integers+BSONValue.swift
nspektor: done

Sources/BSON/Integers+BSONValue.swift
nspektor: okay, I changed the wording a bit to make it a bit clearer that we are showing the keys

Sources/BSON/BSONError.swift
```diff
@@ -42,6 +42,22 @@ public enum BSONError {
...
 
+extension DecodingError {
+    /// Standardize the errors emitted by BSONValue initializers.
+    internal static func _extendedJSONError(
+        keyPath: [String],
+        debugDescription: String
+    ) -> DecodingError {
+        let debugStart = keyPath.joined(separator: ".") +
+            (keyPath == [] ? "" : ": ")
+        print(debugStart + debugDescription)
```
kmahar, line 53: print statement here needs to be removed

patrickfreed: Looks good to me as well, just needs this print removed.

nspektor: done

PR #24, opened 2020-07-20
SWIFT-918 extended json initializers for all non-recursive BSONValues

nspektor:


Sources/BSON/BSONBinary.swift
```diff
@@ -117,6 +117,72 @@ public struct BSONBinary: Equatable, Hashable {
...
+            }
+            guard let base64Str = base64.stringValue else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `base64` from \"\(base64)\", " +
+                        "input must be a base64-encoded (with padding as =) payload as a string"
+                )
+            }
+            guard let subTypeStr = subTypeInput.stringValue,
+                let subTypeInt = UInt8(subTypeStr),
```
patrickfreed, line 165: since the subtype is stored as a hex value, I think we need `UInt8(subTypeStr, radix: 16)`. Otherwise, it may assume it's initializing a decimal value.

Sources/BSON/BSONCode.swift
```diff
@@ -2,6 +2,46 @@ import NIO
...
+     *   - `DecodingError` if `json` is a partial match or is malformed.
+     */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
+        switch json {
+        case let .object(obj):
+            // canonical extended JSON
+            guard let value = obj["$code"] else {
+                return nil
+            }
+            guard obj.count == 1 else {
```
patrickfreed, line 26: this one's actually a little tricky: We also have a `CodeWithScope` type, which has both `$code` and `$scope` keys. If we find that the two keys are exactly `$code` and `$scope`, we should actually just return `nil` here. Otherwise we can throw.

Sources/BSON/BSONDBPointer.swift
```diff
@@ -16,6 +16,71 @@ public struct BSONDBPointer: Equatable, Hashable {
...
+                let ref = dbPointerObj["$ref"],
+                let id = dbPointerObj["$id"] else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected \"$ref\" and \"$id\" keys, " +
+                        "found \(dbPointerObj.keys.count) key(s) within \"$dbPointer\": \(dbPointerObj.keys)"
+                )
+            }
+            guard let refStr = ref.stringValue,
+                let idStr = id.stringValue else {
```
patrickfreed, line 62: so the `$id` portion of this is actually an object id in _extended JSON_, meaning it's a `{ "$oid": <hexstring> }`, not a hex string itself. So instead of doing `stringValue` here, we can re-use the `ObjectId(fromExtJSON:)` initializer.

Sources/BSON/BSONRegularExpression.swift
```diff
@@ -64,6 +64,78 @@ public struct BSONRegularExpression: Equatable, Hashable {
...
+            guard let value = obj["$regularExpression"] else {
+                return nil
+            }
+            guard obj.count == 1 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$regularExpression\" key, found too many keys: \(obj.keys)"
+                )
+            }
+            guard let regexArr = value.arrayValue else {
```
patrickfreed, line 94: the regular expression format is in the format `{"pattern": <string>, "options": <string> }`, not an array.

Sources/BSON/BSONRegularExpression.swift
```diff
@@ -64,6 +64,78 @@ public struct BSONRegularExpression: Equatable, Hashable {
...
+                )
+            }
+            func areOptionsValid(options: String) -> Bool {
+                for option in options {
+                    guard regexOptsMap.keys.contains(option) else {
+                        return false
+                    }
+                }
+                return true
+                // TODO: decide whether we will be validating options
```
patrickfreed, line 124: I think we can just skip on the validation for now.

Sources/BSON/Double+BSONValue.swift
```diff
@@ -1,6 +1,52 @@
...
+     *   - `DecodingError` if `json` is a partial match or is malformed.
+     */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
+        switch json {
+        case let .number(n):
+            // relaxed extended JSON
+            guard let double = Double(exactly: n) else {
+                return nil
+            }
+            self = double
```
patrickfreed, line 25: since `n` is a `Double` already, we can just assign directly without doing the `Double(exactly:)` initializer.

Sources/BSON/Double+BSONValue.swift
```diff
@@ -1,6 +1,52 @@
...
+            guard let value = obj["$numberDouble"] else {
+                return nil
+            }
+            guard obj.count == 1 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$numberDouble\" key, found too many keys: \(obj.keys)"
+                )
+            }
+            guard let str = value.stringValue, let int = Double(str) else {
```
patrickfreed, line 37: since we're dealing with a `Double` here, can we use a different variable name than `int`?

Sources/BSON/Integers+BSONValue.swift
```diff
@@ -31,7 +31,7 @@ extension Int32: BSONValue {
             guard obj.count == 1 else {
                 throw DecodingError._extendedJSONError(
                     keyPath: keyPath,
-                    debugDescription: "Expected only \"$numberInt\" key, found too many keys: \(obj.keys)"
+                    debugDescription: "\"$numberInt\" key, found too many keys: \(obj.keys)"
```
patrickfreed, line 34: I think this `Expected only` was deleted accidentally.

Sources/BSON/BSONRegularExpression.swift
nspektor: I was going back and forth on this, I think the documentation is missing the quotes around "pattern" which confused me. Ill fix this

Sources/BSON/BSONDBPointer.swift
nspektor: okay done

Sources/BSON/BSONRegularExpression.swift
nspektor: removed

Sources/BSON/BSONCode.swift
patrickfreed: I think we want to avoid the implicit dependency on the other initializer here if we can and throw in all error cases. It just so happens that `$code` + `$scope` isn't an error case, so that one we treat differently, but the rest we leave the same.

nspektor: okay sounds good, ill edit

Sources/BSON/BSONDBPointer.swift
patrickfreed: I think we can omit adding to the keyPath here actually, since we're still within the body of a single "value" type so to speak and haven't traversed any real keys. One way to think about it is that `keyPath` tracks the keys that will show up in the decoded BSON, so all the different `$<blah>` won't be there, but something like `"name"` will be.

nspektor: I would still pass in `keyPath` though right? just not add `"$dbPointer`

patrickfreed: Yeah, exactly.

nspektor: done

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -24,5 +59,260 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
+            .to(throwError(errorType: DecodingError.self))
+        expect(try BSONDecimal128(fromExtJSON: ["$numberDecimal": "5.5", "extra": true], keyPath: []))
+            .to(throwError(errorType: DecodingError.self))
+        expect(try BSONDecimal128(fromExtJSON: ["$numberDecimal": JSON.bool(true)], keyPath: ["key", "path"]))
+            .to(throwError(errorType: DecodingError.self))
+    }
+
+    func testBinary() throws {
+        // Success case
+        try expect(try BSONBinary(fromExtJSON: ["$binary": ["base64": "//8=", "subType": "00"]], keyPath: []))
```
kmahar, line 123: related to Patrick's previous comment about needing to use `radix: 16`, it might be a good idea to test a larger subtype here with a different value in hex than in decimal.
all of the pre-defined subtypes have values 0-6 so you'd need to use a "user-defined" one, which has to be something in the range of `0x80` to `UInt8.max` I believe

Sources/BSON/BSONNulls.swift
```diff
@@ -20,6 +42,45 @@ internal struct BSONNull: BSONValue, Equatable {
...
+            // canonical and relaxed extended JSON
+            guard let value = obj["$undefined"] else {
+                return nil
+            }
+            guard obj.count == 1 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$undefined\" key, found too many keys: \(obj.keys)"
+                )
+            }
```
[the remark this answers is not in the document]
nspektor: this was a great idea! I went through and refactored to do this.

Sources/BSON/JSON.swift
```diff
@@ -136,4 +136,33 @@ extension JSON {
...
+    /// - Parameters:
+    ///   - key: a String representing the one key that the initializer is looking for
+    ///   - `keyPath`: an array of `String`s containing the enclosing JSON keys of the current json being passed in.
+    ///                This is used for error messages.
+    /// - Returns: JSON which is the value at the given `key` in `self`
+    ///            or `nil` if `self` is not an `object` or does not contain the given `key`
+    /// - Throws: `DecodingError` if `self` has too many keys
+    internal func onlyHasKey(key: String, keyPath: [String]) throws -> JSON? {
+        guard case let .object(obj) = self else {
+            return nil
```
nspektor, line 153: Not sure what to do in the case that self isn't an object. throw some kind of error? or since in all current usages of this function already check that the json is an object, we can remove this guard?

kmahar: hmm that's a good question.

it seems to me we need to have the guard here right now, even if for no other reason than to extract the `[String: JSON]` as its correct type, right? otherwise you won't be able to do `obj[key]` a few lines down.

maybe this method actually makes more sense as a method to be defined via extension on `Dictionary where Key == String, Value == JSON`. then it would not be possible to call this method on non-objects, as you'd call it on the dictionary you obtain via the switch statements.

alternatively, another option you could consider is to leave this method as-is, and rely on the guard here rather than checking in the individual methods what the type is. that might allow you to remove the switch statements altogether from a number of those implementations where you don't actually use `obj` directly.
it does look like you still use it a fair amount for constructing error messages though, so maybe that wouldn't help much.

nspektor: Oh I like option 3, it would be nice to be able to remove the switch statement entirely for values which only accept object. I can think about ways to construct the error message without `obj` or maybe just leave the whole switch for those ones that use `obj`

Sources/BSON/JSON.swift
```diff
@@ -136,4 +136,33 @@ extension JSON {
...
+/// Helpers
+extension JSON {
+    /// Helper function used in `BSONValue` initializers that take in extended JSON and need to
+    /// check that an object has only 1 specified key.
+    ///
+    /// - Parameters:
+    ///   - key: a String representing the one key that the initializer is looking for
+    ///   - `keyPath`: an array of `String`s containing the enclosing JSON keys of the current json being passed in.
+    ///                This is used for error messages.
+    /// - Returns: JSON which is the value at the given `key` in `self`
```
kmahar, line 148: update to include what the second returned value is?

nspektor: okay, I changed it but am open to suggestions to make the wording clearer

PR #25, opened 2020-07-23
SWIFT-919 Extended JSON to BSON Conversion for all BSONValues

nspektor:
Recursive bson inits + date

### DIFF
--- a/Sources/BSON/BSON.swift
+++ b/Sources/BSON/BSON.swift
@@ -105,11 +105,10 @@ public enum BSON {
             }
         }
         // Document accepts any JSON object so it should be tried after all the more specific BSON types are tried
-        if let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) {
-            self = doc.bson
-            return
+        guard let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) else {
+            throw BSONError.InternalError(message: "Could not parse BSON from \(json)")
         }
-        throw BSONError.InternalError(message: "Could not parse BSON from \(json)")
+        self = doc.bson
     }
 
     /// Get the `BSONType` of this `BSON`.
--- a/Sources/BSON/BSON.swift
+++ b/Sources/BSON/BSON.swift
@@ -96,7 +96,7 @@ public enum BSON {
             return
         }
         for bsonType in BSON.allBSONTypes.values {
-            if bsonType == Int32.self || bsonType == Int64.self || bsonType == BSONDocument.self {
+            guard bsonType != Int32.self && bsonType != Int64.self && bsonType != BSONDocument.self else {
                 continue
             }
             if let value = try bsonType.init(fromExtJSON: json, keyPath: keyPath) {
--- a/Sources/BSON/BSON.swift
+++ b/Sources/BSON/BSON.swift
@@ -84,7 +84,7 @@ public enum BSON {
     ///              This is used for error messages.
     ///
     /// Throws:
-    ///   - `InternalError`
+    ///   - `DecodingError` if `json` is malformatted extended json
     internal init(fromExtJSON json: JSON, keyPath: [String]) throws {
         // Spec requires that we try int32, try int64, then try double for decoding numbers
         if let int32 = try Int32(fromExtJSON: json, keyPath: keyPath) {
@@ -106,7 +106,6 @@ public enum BSON {
         }
         // Document accepts any JSON object so it should be tried after all the more specific BSON types are tried
         if let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) {
-            print(doc)
             self = doc.bson
             return
         }
--- a/Sources/BSON/BSON.swift
+++ b/Sources/BSON/BSON.swift
@@ -77,6 +77,42 @@ public enum BSON {
         }
     }
 
+    /// Initialize a `BSON` from ExtendedJSON
+    /// Parameters:
+    ///   - `json`: a `JSON` representing the canonical or relaxed form of ExtendedJSON for any `BSONValue`.
+    ///   - `keyPath`: an array of `Strings`s containing the enclosing JSON keys of the current json being passed in.
+    ///              This is used for error messages.
+    ///
+    /// Throws:
+    ///   - `InternalError`
+    internal init(fromExtJSON json: JSON, keyPath: [String]) throws {
+        // Spec requires that we try int32, try int64, then try double for decoding numbers
+        if let int32 = try Int32(fromExtJSON: json, keyPath: keyPath) {
+            self = int32.bson
+            return
+        }
+        if let int64 = try Int64(fromExtJSON: json, keyPath: keyPath) {
+            self = int64.bson
+            return
+        }
+        for bsonType in BSON.allBSONTypes.values {
+            if bsonType == Int32.self || bsonType == Int64.self || bsonType == BSONDocument.self {
+                continue
+            }
+            if let value = try bsonType.init(fromExtJSON: json, keyPath: keyPath) {
+                self = value.bson
+                return
+            }
+        }
+        // Document accepts any JSON object so it should be tried after all the more specific BSON types are tried
+        if let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) {
+            print(doc)
+            self = doc.bson
+            return
+        }
+        throw BSONError.InternalError(message: "Could not parse BSON from \(json)")
+    }
+
     /// Get the `BSONType` of this `BSON`.
     public var type: BSONType {
         self.bsonValue.bsonType
--- a/Sources/BSON/BSONCode.swift
+++ b/Sources/BSON/BSONCode.swift
@@ -45,7 +45,7 @@ extension BSONCode: BSONValue {
     internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
         switch json {
         case let .object(obj):
-            // canonical extended JSON
+            // canonical and relaxed extended JSON
             guard let value = obj["$code"] else {
                 return nil
             }
@@ -88,6 +88,54 @@ extension BSONCode: BSONValue {
 }
 
 extension BSONCodeWithScope: BSONValue {
+    /*
+     * Initializes a `BSONCode` from ExtendedJSON.
+     *
+     * Parameters:
+     *   - `json`: a `JSON` representing the canonical or relaxed form of ExtendedJSON for `Code`.
+     *   - `keyPath`: an array of `String`s containing the enclosing JSON keys of the current json being passed in.
+     *              This is used for error messages.
+     *
+     * Returns:
+     *   - `nil` if the provided value is not a `String`.
+     *
+     * Throws:
+     *   - `DecodingError` if `json` is a partial match or is malformed.
+     */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
+        switch json {
+        case let .object(obj):
+            // canonical and relaxed extended JSON
+            guard
+                let code = obj["$code"],
+                let scope = obj["$scope"]
+            else {
+                return nil
+            }
+            guard obj.count == 2 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$code\" and \"$scope\" keys, got: \(obj.keys)"
+                )
+            }
+            guard let codeStr = code.stringValue else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `BSONCode` \"\(code)\", input must be a string."
+                )
+            }
+            guard let scopeDoc = try BSONDocument(fromExtJSON: scope, keyPath: keyPath + [codeStr]) else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse scope from \"\(scope)\", input must be a Document."
+                )
+            }
+            self = BSONCodeWithScope(code: codeStr, scope: scopeDoc)
+        default:
+            return nil
+        }
+    }
+
     internal static var bsonType: BSONType { .codeWithScope }
 
     internal var bson: BSON { .codeWithScope(self) }
--- a/Sources/BSON/BSONCode.swift
+++ b/Sources/BSON/BSONCode.swift
@@ -106,17 +106,8 @@ extension BSONCodeWithScope: BSONValue {
         switch json {
         case let .object(obj):
             // canonical and relaxed extended JSON
-            guard
-                let code = obj["$code"],
-                let scope = obj["$scope"]
-            else {
+            guard let (code, scope) = try json.unwrapObject(withKeys: "$code", "$scope", keyPath: keyPath) else {
                 return nil
-            }
-            guard obj.count == 2 else {
-                throw DecodingError._extendedJSONError(
-                    keyPath: keyPath,
-                    debugDescription: "Expected only \"$code\" and \"$scope\" keys, got: \(obj.keys)"
-                )
             }
             guard let codeStr = code.stringValue else {
                 throw DecodingError._extendedJSONError(
@@ -125,7 +116,7 @@ extension BSONCodeWithScope: BSONValue {
                         " input must be a string."
                 )
             }
-            guard let scopeDoc = try BSONDocument(fromExtJSON: scope, keyPath: keyPath + [codeStr]) else {
+            guard let scopeDoc = try BSONDocument(fromExtJSON: scope, keyPath: keyPath + ["$scope"]) else {
                 throw DecodingError._extendedJSONError(
                     keyPath: keyPath,
                     debugDescription: "Could not parse scope from \"\(scope)\", input must be a Document."
--- a/Sources/BSON/BSONCode.swift
+++ b/Sources/BSON/BSONCode.swift
@@ -121,7 +121,8 @@ extension BSONCodeWithScope: BSONValue {
             guard let codeStr = code.stringValue else {
                 throw DecodingError._extendedJSONError(
                     keyPath: keyPath,
-                    debugDescription: "Could not parse `BSONCode` \"\(code)\", input must be a string."
+                    debugDescription: "Could not parse `BSONCodeWithScope` from \"code\": \"\(code)\"," +
+                        " input must be a string."
                 )
             }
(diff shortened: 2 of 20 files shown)


Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -157,11 +165,33 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
     }
 
     func testCodeWScope() throws {
-        // Skipping until there is a BSON.init(fromExtJSON) because its recursive
+        // Success case
+        expect(try BSONCodeWithScope(fromExtJSON: ["$code": "javascript", "$scope": ["doc": "scope"]], keyPath: []))
+            .to(equal(BSONCodeWithScope(
+                code: "javascript",
+                scope: BSONDocument(keyValuePairs: [("doc", BSON.string("scope"))])
```
patrickfreed, line 172: you can initialize this directly from an object literal like you do with `JSON` (e.g. `["doc": "scope"]`)

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -251,6 +281,15 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
             .to(throwError(errorType: DecodingError.self))
     }
 
+    func testDatetime() throws {
+        // Canonical Success case
+        expect(try Date(fromExtJSON: ["$date": ["$numberLong": "500004"]], keyPath: []))
+            .to(equal(Date(msSinceEpoch: 500_004)))
+        // Relaxed Success case
+        expect(try Date(fromExtJSON: ["$date": "2012-12-24T12:15:30.501Z"], keyPath: []))
+            .to(equal(ExtendedJSONDecoder.extJSONDateFormatter.date(from: "2012-12-24T12:15:30.501Z")))
```
patrickfreed, line 290: this will basically always pass because its relying on the same formatter for parsing that the code is using. To make this test a little more meaningful, we should derive the expected date from another means like the milliseconds since epoch (which in this case for this date is `1356351330501`).

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -297,7 +336,18 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
     }
 
     func testArray() throws {
-        // Skipping until there is a BSON.init(fromExtJSON) because its recursive
+        // Success cases
+        expect(try Array(fromExtJSON: [1, ["$numberLong": "2"], "3"], keyPath: []))
+            .to(equal([BSON.int32(Int32(1)), BSON.int64(Int64(2)), BSON.string("3")]))
```
patrickfreed, line 341: the compiler will be able to infer that 1 and 2 here are `Int32` and `Int64` respectively by nature of being included in a `BSON.int32` invocation, so the casts can be omitted.

Sources/BSON/Date+BSONValue.swift
```diff
@@ -2,6 +2,51 @@ import Foundation
...
+     *   - `DecodingError` if `json` is a partial match or is malformed.
+     */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
+        guard let (value, _) = try json.isObjectWithSingleKey(key: "$date", keyPath: keyPath) else {
+            return nil
+        }
+        switch value {
+        case let .object(obj):
+            // canonical extended JSON
+            print(obj)
```
patrickfreed, line 26: debug print

Sources/BSON/Date+BSONValue.swift
```diff
@@ -2,6 +2,51 @@ import Foundation
...
+            guard let int = try Int64(fromExtJSON: value, keyPath: keyPath + ["$date"]) else {
```
patrickfreed, line 27: so we dont wan't to be adding `$date` to the keypath, as its one of those reserved ones that won't show up in the resultant BSON.

Sources/BSON/BSONCode.swift
```diff
@@ -88,6 +88,54 @@ extension BSONCode: BSONValue {
...
+            guard obj.count == 2 else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Expected only \"$code\" and \"$scope\" keys, got: \(obj.keys)"
+                )
+            }
+            guard let codeStr = code.stringValue else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `BSONCode` \"\(code)\", input must be a string."
```
patrickfreed, line 124: `BSONCode` -> `BSONCodeWithScope`

Sources/BSON/BSON.swift
```diff
@@ -77,6 +77,42 @@ public enum BSON {
...
     }
 
+    /// Initialize a `BSON` from ExtendedJSON
+    /// Parameters:
+    ///   - `json`: a `JSON` representing the canonical or relaxed form of ExtendedJSON for any `BSONValue`.
+    ///   - `keyPath`: an array of `Strings`s containing the enclosing JSON keys of the current json being passed in.
+    ///              This is used for error messages.
+    ///
+    /// Throws:
+    ///   - `InternalError`
```
patrickfreed, line 87: Generally we don't call out when `InternalError`s are thrown, since they theoretically shouldn't ever occur. We should call out that this will throw `DecodingError`s when malformatted extjson is passed in though.

Sources/BSON/BSON.swift
```diff
@@ -77,6 +77,42 @@ public enum BSON {
...
+                continue
+            }
+            if let value = try bsonType.init(fromExtJSON: json, keyPath: keyPath) {
+                self = value.bson
+                return
+            }
+        }
+        // Document accepts any JSON object so it should be tried after all the more specific BSON types are tried
+        if let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) {
+            print(doc)
```
patrickfreed, line 109: debug print

Sources/BSON/BSON.swift
```diff
@@ -77,6 +77,42 @@ public enum BSON {
...
+            if bsonType == Int32.self || bsonType == Int64.self || bsonType == BSONDocument.self {
+                continue
+            }
+            if let value = try bsonType.init(fromExtJSON: json, keyPath: keyPath) {
+                self = value.bson
+                return
+            }
+        }
+        // Document accepts any JSON object so it should be tried after all the more specific BSON types are tried
+        if let doc = try BSONDocument(fromExtJSON: json, keyPath: keyPath) {
```
patrickfreed, line 108: since this is something that should happen, can we invert this to be a `guard` instead?

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -157,11 +165,33 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
-        // Skipping until there is a BSON.init(fromExtJSON) because its recursive
+        // Success case
+        expect(try BSONDocument(fromExtJSON: ["key": ["$numberInt": "5"]], keyPath: []))
+            .to(equal(BSONDocument(keyValuePairs: [("key", .int32(5))])))
+        // Nil case
+        expect(try BSONDocument(fromExtJSON: 1, keyPath: [])).to(beNil())
+
+        // Error case
+        expect { try BSONDocument(fromExtJSON: ["time": ["$timestamp": 5]], keyPath: []) }
+            .to(throwError(DecodingError._extendedJSONError(
```
patrickfreed, line 191: This test is a little too specific to the implementation, which means it will need to be updated if we ever change the format of the error message. The key things we do need to test here are that the error is in fact a `DecodingError` and that the keypath shows up properly. To do so, you can use that nimble matcher I showed you and do some checks on the error:

```swift
expect(try BSONDocument(...)).to(throwError { error in
    expect(error).to(matchError(DecodingError.self))
    expect(error.localizedDescription).to(contain("time"))
})
```
This is a little more resilient to implementation changes while still verifying the behavior we want.

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -5,6 +5,14 @@ import NIO
 import XCTest
 
 open class ExtendedJSONConversionTestCase: BSONTestCase {
+    func testAnyExtJSON() throws {
+        // Success cases
+        expect(try BSON(fromExtJSON: "hello", keyPath: [])).to(equal(BSON.string("hello")))
+        let document = try BSON(fromExtJSON: ["num": ["$numberInt": "5"], "extra": 1], keyPath: [])
+        expect(document.documentValue!["num"]).to(equal(.int32(5)))
```
patrickfreed, line 12: these `!` casts will crash the test runner if they fail, so it's good practice to include a `expect(document.documentValue).toNot(beNil())` before them to have the test fail more naturally.

Sources/BSON/BSONValue.swift
```diff
@@ -18,7 +18,7 @@ internal protocol BSONValue: Codable {
     /// Initializes a corresponding `BSON` from the provided extendedJSON.
```
patrickfreed, line 18: this docstring can be updated now that this method is fully implemented.

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: I tried using exactly the code you suggested, but `error.localizedDescription` returns "The data couldn’t be read because it isn’t in the correct format." somehow I need to get at the `debugDescription` in the context of `DecodingError.dataCorrupted`, but I cant figure out a way to do so.

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: ok

Sources/BSON/BSON.swift
nspektor: yep

Tests/BSONTests/ExtendedJSONConversionTests.swift
patrickfreed: Discussed offline: can use `String(describing:)` here.

Tests/BSONTests/ExtendedJSONConversionTests.swift
patrickfreed: discussed offline: it's fine to use `!` in tests after asserting non-nil

Sources/BSON/BSONValue.swift
patrickfreed: So this initializer is initializing an instance of the `BSONValue` itself, not a `BSON` (like what `read` is doing). So could we update the docstring to mention that?

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -157,11 +166,33 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
+            .to(throwError(errorType: DecodingError.self))
+        expect(try BSONCodeWithScope(fromExtJSON: ["$code": "javascript", "$scope": 1], keyPath: []))
+            .to(throwError(errorType: DecodingError.self))
     }
 
     func testDocument() throws {
-        // Skipping until there is a BSON.init(fromExtJSON) because its recursive
+        // Success case
+        expect(try BSONDocument(fromExtJSON: ["key": ["$numberInt": "5"]], keyPath: []))
+            .to(equal(BSONDocument(keyValuePairs: [("key", .int32(5))])))
```
patrickfreed, line 186: ditto use document literal here

Sources/BSON/BSONValue.swift
nspektor: yep

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

Sources/BSON/Date+BSONValue.swift
```diff
@@ -2,6 +2,50 @@ import Foundation
...
+     *              This is used for error messages.
+     *
+     * Returns:
+     *   - `nil` if the provided value does not conform to the `Date` syntax.
+     *
+     * Throws:
+     *   - `DecodingError` if `json` is a partial match or is malformed.
+     */
+    internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
+        guard let (value, _) = try json.isObjectWithSingleKey(key: "$date", keyPath: keyPath) else {
```
patrickfreed, line 20: I know this was from the last review, but I just noticed now that we generally don't use the second part of this tuple, and in the places we do, I'm not sure we need to. For instance, in `BSONRegularExpression.swift`, we use it to print the keys in the error message, but the keys we actually want are those of the interior `regexObj`, not of `obj`. By nature of the method succeeding and returning non-nil, we know that `obj.keys` is `["$regularExpression"]` in that case. The same applies for `BSONBinary`.

Also, since this method is retrieving something, I think we should use a prefix that indicates that, since `is` is typically reserved for checking a condition and returning a boolean value.

Finally, to address the situations where this needs to be used with two different values (e.g. the interior object for `BSONRegularExpression`), we could use a similar method to reduce boilerplate. This doesn't scale well, but we only need it for exactly one or two values, so I think we can implement a method for this in an ad-hoc manner. Also, there is precedent for implementing methods like this in [Swift for TensorFlow](https://www.tensorflow.org/swift/api_docs/Protocols/Differentiable#sequencedthrough:_:) (see the various `sequenced` overloads).

To address these concerns, I suggest refactoring this method and also introducing a new one for two values:
```swift
extension JSON {
    func unwrapObject(withKeys: String, keyPath: [String]) throws -> (JSON)? {}
    func unwrapObject(withKeys key1: String, _ key2: String, keyPath: [String]) throws -> (JSON, JSON)? {}
}
```

Then usage would look like:
```swift
guard let (regexValue) = try json.unwrapObject(withKeys: "$regularExpression") else { ... }
guard let (pattern, options) = try regexValue.unwrapObject(withKeys: "pattern", "options") else { ... }
```

Sources/BSON/BSONCode.swift
```diff
@@ -88,6 +88,55 @@ extension BSONCode: BSONValue {
...
+                )
+            }
+            guard let codeStr = code.stringValue else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `BSONCodeWithScope` from \"code\": \"\(code)\"," +
+                        " input must be a string."
+                )
+            }
+            guard let scopeDoc = try BSONDocument(fromExtJSON: scope, keyPath: keyPath + [codeStr]) else {
```
patrickfreed, line 128: `keyPath` is reserved for keys only, so we don't want to be appending the code value here.

In general, the keyPath just tracks where in the output BSON document we are when we encounter an error in the input JSON. This allows us to tell the user where to look for an error occurring.

For example, given input JSON:
```json
{ 
    "foo": { "$numberInt": "12" },
    "bar": { "baz": { "$code": "console.log(x)", "$scope": { "x": 5 } } }
}
```
the output BSON document would look like:
```javascript
{
    "foo": 12,
    "bar": { "baz": CodeWithScope(...) }
}
```
So the only values that should ever end up in keyPath are `["foo"]` or `["bar"]` or `["bar, "baz"]`, since those are the only _actual keys_. All the other keys encountered ("$code", "$scope", etc.) are part of the _values_ themselves.

Now, if the value for the `CodeWithScope` in this example was malformatted and we threw an error (say the scope was incorrect), `keyPath` would be `["bar", "baz"]` and our error would then include the string "bar.baz", so they know what the issue was. 

So, getting back to the original comment, `codeStr` here refers to `"console.log(x)"` in this example, so we definitely don't want to be putting that in the keyPath. More specifically, we only ever need to modify the keyPath when descending recursively from within a `BSONDocument` or `[BSON]` initializer.

patrickfreed: Correction: I realize now that when we recursively enter the scope `BSONDocument`, we actually do need to modify the keyPath in the event that it has a decoding error somewhere down the line within the scope document. This is probably what you were going for originally.

For example, for the following badly formatted JSON:
```javascript
{ "mycode": { "$code": "console.log(5)", "$scope": { "x": 2, "y": { "$numberInt": "ERROR" } } } }
````
we don't want this to give us `DecodingError("mycode.y: ...")`, we want it to say `DecodingError("mycode.$scope.y: ...")`.

So, most of what I said still holds, but you are correct that we need to append to the keyPath here--just instead of appending `codeStr` we append `"$scope"`. To generalize this rule: we append to the keypath whenever we recurse via `BSON`, `Array`, or `BSONDocument`. 

Sorry for the confusion!

File: Tests/BSONTests/ExtendedJSONConversionTests.swift [hunk 16]
```diff
@@ -157,11 +166,33 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
-        // Skipping until there is a BSON.init(fromExtJSON) because its recursive
+        // Success case
+        expect(try BSONDocument(fromExtJSON: ["key": ["$numberInt": "5"]], keyPath: []))
+            .to(equal(["key": .int32(5)]))
+        // Nil case
+        expect(try BSONDocument(fromExtJSON: 1, keyPath: [])).to(beNil())
+
+        // Error case
+        expect(try BSONDocument(fromExtJSON: ["foo": ["bar": ["$timestamp": 5]]], keyPath: []))
+            .to(throwError { error in
```
kmahar, line 192: TIL you can use a closure here! nice

Sources/BSON/BSON.swift
```diff
@@ -77,6 +77,40 @@ public enum BSON {
...
+        if let int32 = try Int32(fromExtJSON: json, keyPath: keyPath) {
+            self = int32.bson
+            return
+        }
+        if let int64 = try Int64(fromExtJSON: json, keyPath: keyPath) {
+            self = int64.bson
+            return
+        }
+        for bsonType in BSON.allBSONTypes.values {
+            if bsonType == Int32.self || bsonType == Int64.self || bsonType == BSONDocument.self {
```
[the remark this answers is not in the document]
nspektor: I was having some issues with the way you suggested because the `BSONValue`s don't conform to Equatable. Patrick and I discussed this and decided to leave it with the guard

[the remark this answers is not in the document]
kmahar: sgtm

PR #26, opened 2020-07-24
SWIFT-920 Support Decoding from BSON in BSONDecoder

nspektor:


Sources/BSON/BSONDecoder.swift
```diff
@@ -165,6 +154,33 @@ public class BSONDecoder {
...
+            return b
+        }
+        let _decoder = _BSONDecoder(referencing: bson, options: self.options)
+        do {
+            return try type.init(from: _decoder)
+        } catch let error as BSONErrorProtocol {
+            let unknownErrorMessage = "Unknown Error occurred while decoding BSON"
+            throw DecodingError.dataCorrupted(
+                DecodingError.Context(
+                    codingPath: [],
```
kmahar, line 177: I know this is a copy paste, but now that I am looking at it I think it would make sense to use `_decoder.codingPath` here. if we encounter a BSON error it did not necessarily occur at the top level.

Sources/BSON/BSONDecoder.swift
```diff
@@ -165,6 +154,33 @@ public class BSONDecoder {
...
+     * Decodes a top-level value of the given type from the given BSON.
+     *
+     * - Parameter type: The type of the value to decode.
+     * - Parameter document: The BSON to decode from.
+     * - Returns: A value of the requested type.
+     * - Throws: `DecodingError` if any value throws an error during decoding.
+     */
+    internal func decode<T: Decodable>(_ type: T.Type, fromBSON bson: BSON) throws -> T {
+        // if the requested type is `BSON` we're done
+        if let b = bson as? T {
```
kmahar, line 167: if I remember correctly, we are using this method for the final conversion step right, from BSON to the user's requested type?
if so, it might make sense to also special-case when `bson.bsonValue` is an instance of `T`. 

in fact, I might be wrong but we actually might have to do that, because for all of our custom BSON types (e.g. `Binary`) we just throw an error if you actually try to use the generalized decoding implementation (called a few lines down), as those types (for now anyway) only are meant to work with encoders and decoders we own (BSON and now extJSON):
https://github.com/mongodb/swift-bson/blob/e0b0efbd7ea1175559a8e4f3cdf3d44de4813632/Sources/BSON/BSONValue.swift#L35

patrickfreed: Oh, nice catch. I think what would be even better would be to utilize the same code path that "interior" decoding would use by invoking `unbox` directly instead of `Type.init`. This is what [`JSONDecoder` does](https://github.com/apple/swift-corelibs-foundation/blob/master/Sources/Foundation/JSONEncoder.swift#L1163), for instance.

patrickfreed: `unbox` also has special handling if `T` is a `BSON` as well, so this `if let` can go away too.

Sources/BSON/BSONDecoder.swift
nspektor: ok

Sources/BSON/BSONDecoder.swift
nspektor: I think the change I just made should address this

Tests/BSONTests/CodecTests.swift
```diff
@@ -41,6 +41,14 @@ final class CodecTests: BSONTestCase {
         let s: NestedStruct
     }
 
+    /// Test decoding non-document BSON
+    func testAnyBSON() throws {
```
patrickfreed, line 45: could we add a case here that uses one of the types introduced in the library (e.g. `ObjectId`)? This will verify whether the changes above were adequate.

nspektor: added

nspektor:
https://spruce.mongodb.com/version/5f1b5416d6d80a01691cf840/tasks - evergreen link since I guess it didnt get run automatically

PR #27, opened 2020-07-24
SWIFT-921 implement extended json decoder methods

nspektor:


Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -5,6 +5,18 @@ import NIO
 import XCTest
 
 open class ExtendedJSONConversionTestCase: BSONTestCase {
+    func testExtendedJSONDecoder() throws {
+        struct Foo: Decodable, Equatable {
```
kmahar, line 9: to test my theory mentioned in the other PR about whether we need to special-case, you could add one of the BSON types we own like Binary or regex to this example.

nspektor: added :)

kmahar: I realized my comment here isn't correct, since what came up in the other PR related to the top-level API rather than what happens for nested types. that said, I think the test you added there using ObjectID accomplishes what I was trying to get at here anyway.

PR #28, opened 2020-07-24
Hodgepodge of test fixes

kmahar:
As I mentioned in the team channel I noticed we had some tests that were inadvertently left disabled. I've now enabled all of them and made fixes as needed to both the driver to get stuff to pass as well as to the tests to bring them in line with the new API.

FYI I have broken this into two commits (the first is just moving files) - I recommend using the filters to review the changes from the second commit only. for whatever reason that view recognizes I didn't make totally new files and just moved + edited existing ones, but when you look at the PR diff as a whole it considers the moved + edited files to be new files 🤔🤔🤔


kmahar:
Link for your convenience to only review the second commit: https://github.com/mongodb/swift-bson/pull/28/commits/1d79fbe826cbe070012fe4575c3a6f6108922930

kmahar:
Also, just to provide some more context for @nspektor : at the start of the BSON project we copied over all of the existing tests from the driver's built-in BSON library and put them in a "disabled tests" folder that `swift test` would ignore. throughout the project as we implemented more API, we slowly enabled more of them, but some got left behind.

a lot of the changes you see here are also just renaming types - we copied the files over here and then later on decided to prefix all of our types with `BSON`.

Tests/BSONTests/DocumentTests.swift
```diff
@@ -252,60 +233,43 @@ final class DocumentTests: MongoSwiftTestCase {
...
-        expect(doc).to(equal(fromRawBSON))
-    }
+    // TODO: SWIFT-930: unskip
+    // func testRawBSON() throws {
+    //     let doc = try BSONDocument(fromJSON: "{\"a\" : [{\"$numberInt\": \"10\"}]}")
+    //     let fromRawBSON = try BSONDocument(fromBSON: doc.rawBSON)
+    //     expect(doc).to(equal(fromRawBSON))
+    // }
 
     func testCopyOnWriteBehavior() {
```
kmahar, line 247: this test has been largely simplified, as we can now assume that ByteBuffer correctly handles sharing storage for us. in the past we implemented copy-on-write ourselves so this test was checking that storage would be correctly copied on mutations.

Tests/BSONTests/DocumentTests.swift
```diff
@@ -345,249 +296,121 @@ final class DocumentTests: MongoSwiftTestCase {
...
-    func testOverwritable() throws {
-        // make a deep copy so we start off with uniquely referenced storage
-        var doc = DocumentTests.overwritables.withBSONPointer { ptr in
-            Document(copying: ptr)
-        }
-
-        // save a reference to original bson_t so we can verify it doesn't change
-        let pointer = doc.pointerAddress
+    // Test overwriting with value of same type
+    func testOverwritingValues() throws {
```
kmahar, line 300: previously we followed a separate code path when overwriting values of fixed lengths (e.g. ObjectID) because libbson had special API to do that in-place. we don't do that now (though we could revisit it as a future optimization) so I've just turned the tests for all that overwriting-in-place behavior into general tests for overwriting behavior since that still seemed useful to have.

Tests/BSONTests/DocumentTests.swift
```diff
@@ -345,249 +296,121 @@ final class DocumentTests: MongoSwiftTestCase {
...
-        ["double", "int32", "int64", "bool", "decimal", "oid", "timestamp", "datetime"].forEach {
-            overwritableDoc[$0] = .null
-            // the storage should change every time
-            expect(overwritableDoc.pointerAddress).toNot(equal(overwritablePointer))
-            overwritablePointer = overwritableDoc.pointerAddress
-        }
-
-        var nonOverwritableDoc = DocumentTests.nonOverwritables.withBSONPointer { ptr in
-            Document(copying: ptr)
+    func testDeletions() throws {
```
kmahar, line 404: this was replacing everything with `null`, but I think the intent of the test was really probably to test deletions, and it got garbled when we did a project to make it so users could differentiate between null and nil. so I've switched it to test deletions instead

Tests/BSONTests/DocumentTests.swift
```diff
@@ -892,7 +715,8 @@ final class DocumentTests: MongoSwiftTestCase {
             try container.encode(customEncodedDoc)
         }
         decoder.dataDecodingStrategy = .custom { decoder in
-            let doc = try Document(from: decoder)
+            let container = try decoder.singleValueContainer()
```
kmahar, line 718: @patrickfreed, this is related to our conversation on a recent PR: https://github.com/mongodb/swift-bson/pull/19#discussion_r451186379

since this code was directly calling through to `Document`'s `Codable` conformance (even though it was only ever happening with a `BSONDecoder`), we would end up calling the default `BSONValue.init(from decoder: Decoder)` implementation (which just calls `getDecodingError`, since we assumed no one would ever hit it while using a `BSONDecoder`) and then getting a decoding error. 

very confusingly, the error here was a type mismatch where both the expected and actual type were the same (`BSONDocument`), because we would hit this logic: https://github.com/mongodb/swift-bson/blob/aa44622750207f305dbd16769adae552255ece65/Sources/BSON/BSONError.swift#L140

this makes me think that maybe we need to revisit the idea of just erroring in that method and maybe need to check for decoder type, because it's possible even when using a decoder that we own to hit such an error.

I've updated to this workaround for now which goes through the decoder's container API and therefore hits our special-casing.

though, thinking about it more, I actually think it's unwise to use these initializers in general, because they can inadvertently bypass decoding strategies where the default implementation of the method is not used at all. 
for example, the way we implement our `Date` decoding strategies (and the way I believe `JSONDecoder` does it too) relies on seeing a call to decode a `Date` from a decoding container. We then switch according to the strategy, and only actually use that method at all in the `.deferredToDate` case:
https://github.com/mongodb/swift-bson/blob/aa44622750207f305dbd16769adae552255ece65/Sources/BSON/BSONDecoder.swift#L400

while the `encode(to: encoder)` and `init(from: decoder)` methods are public, I think they're actually not really intended for consumption except for within encoders and decoders.

patrickfreed: Nice catch, and I think you're right about avoiding direct usage of the `Codable` requirements, so I don't think what you've done is really a workaround but in fact is just the right way to do it.

Sources/BSON/BSON.swift
```diff
@@ -302,6 +302,37 @@ extension BSON {
             return nil
         }
     }
+
+    /// Return this BSON as a `Double` if possible.
+    /// This will coerce numeric cases (e.g. `.decimal128`) into a `Double` if such coercion would be lossless.
+    public func toDouble() -> Double? {
```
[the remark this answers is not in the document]
kmahar: hmm yeah good point. this is just what was in the driver already. how do you think we should handle these?
for some conversions (e.g. Decimal128 to double) I think it's actually pretty straightforward because it looks like the Double initializer accepting a string will accept Decimal128 strings (e.g. "1.2E10" is accepted). but I don't believe the Int initializer accepts those, though I suppose we could maybe funnel the conversion through Double, like Int(Double(decimal128Str))?

alternatively we could just say these methods don't work with Decimal128s.

cc @nbbeeken

nspektor:
lgtm 🎉

Tests/BSONTests/DocumentTests.swift
kmahar: that said, although it is unwise to do this, it's an easy mistake to make, and the error we currently output is incorrect / unhelpful. 

I'm wondering if we should handle this in the default conformances on BSONValue like:
```swift
    public init(from decoder: Decoder) throws {
        guard let bsonDecoder = decoder as? _BSONDecoder else {
            throw getDecodingError(type: Self.self, decoder: decoder)
        }

        let topContainer = bsonDecoder.storage.topContainer
        guard let value = topContainer.bsonValue as? Self else {
            throw DecodingError._typeMismatch(at: decoder.codingPath, expectation: Self.self, reality: topContainer.bsonValue)
        }
        self = value
    }
```

```swift
    public func encode(to encoder: Encoder) throws {
        guard let bsonEncoder = encoder as? _BSONEncoder else {
            throw bsonEncodingUnsupportedError(value: self, at: encoder.codingPath)
        }
        bsonEncoder.storage.containers.append(self)
    }
```

what do you think?

alternatively we could just throw a different error message if it's a BSON/extJSON encoder/decoder vs any other kind.

Sources/BSON/BSON.swift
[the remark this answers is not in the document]
nbbeeken: Just giving it a whirl in the swift shell to see the results, this string is a valid Dec128 number
```swift
let d = Double("1.234567890123456789012345678901234E-7") 
// d: Double? = 1.2345678901234568E-7 
let i = Int(d!)
// i: Int = 0
```
So we could go with that method but it seems like loss of precision is not a typical use case for decimal 128, I'd vote for going with they don't work with d128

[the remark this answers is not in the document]
patrickfreed: You can avoid the lossy conversions by doing using the `Int(exactly:)` initializers. That being said, I think I agree that decimal128 probably shouldn't interface with these methods since it really is a different beast that needs to be handled carefully. For example, its definition of "lossy" is different than that of regular numbers, e.g. `BSONDecimal128("5") != BSONDecimal128("5.0")` (at least according to the existing tests).

That being said, we do have the existing `toDecimal128` method, so it may be confusing if certain numbers map _to_ decimal128 but decimal128 doesnt map back to them. I honestly am not really bothered by this, but I figured it was worth pointing out. I have a feeling that `toDecimal128` will be rarely if ever used in practice.

Tests/BSONTests/DocumentTests.swift
patrickfreed: Yeah this seems like a good approach.

Sources/BSON/BSON.swift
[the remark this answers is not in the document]
patrickfreed: Just to circle back on this--are we okay with just updating the docstring to remove the reference to decimal128?

Tests/BSONTests/DocumentTests.swift
patrickfreed: Is this something we should do here, or as part of follow on work?

Sources/BSON/BSON.swift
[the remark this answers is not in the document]
kmahar: whoops, sorry I haven't had time to get back to this PR yet and make the changes around the default Codable conformance for BSONValues. that's fine with me though. did this come up elsewhere?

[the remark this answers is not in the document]
patrickfreed: Nope, just wanted to check if you were waiting on me for anything.

Tests/BSONTests/DocumentTests.swift
kmahar: I opened [SWIFT-967](https://jira.mongodb.org/browse/SWIFT-967).

Sources/BSON/BSON.swift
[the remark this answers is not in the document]
kmahar: updated the docstring.

PR #29, opened 2020-07-27
SWIFT-922 Sketch of public API for ExtendedJSONEncoder

nspektor:


Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+/// Facilitates the encoding of `Encodable` values into ExtendedJSON.
+public class ExtendedJSONEncoder {
+    internal enum extendedJSONMode {
+        case canonical
+        case relaxed
+    }
```
nspektor, line 7: Not sure If adding this enum is technically part of sketching the public API

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    /// Initialize an `ExtendedJSONEncoder`.
+    public init() {
```
nspektor, line 10: This initializer will eventually be taking in options right? Should I add something as part of this PR to include that?

Sources/BSON/ExtendedJSONEncoder.swift
patrickfreed: discussed offline: this is part of the public API--it's the enum users can set to determine whether the encoder produces relaxed or canonical extended JSON

Sources/BSON/ExtendedJSONEncoder.swift
patrickfreed: discussed offline: yep, add options to this

nspektor: done

Sources/BSON/ExtendedJSONEncoder.swift
nspektor: done

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
```
kmahar, line 1: nit: add a line break between the import statement and the docstring

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,39 @@
...
+/// Facilitates the encoding of `Encodable` values into ExtendedJSON.
+public class ExtendedJSONEncoder {
+    /// The enum representing the two possible encoding modes.
```
kmahar, line 4: I think it would be helpful to include some more info the docstrings here about the different between canonical and relaxed. you could take some inspiration from the definitions in the extJSON spec:
https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#terms

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,39 @@
...
+    public enum extendedJSONMode {
```
kmahar, line 5: I think we could just name this `Mode`, making the full type name `ExtendedJSONEncoder.Mode`.

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /// The options set on the encoder.
+    public struct Options {
+        /// Either Canonical or Relaxed Extended JSON Format to encode to.
+        public var mode: extendedJSONMode
+    }
+
+    /// Initialize an `ExtendedJSONEncoder`.
+    public init(options: ExtendedJSONEncoder.Options) {
```
kmahar, line 20: We talked about this on the design doc, but I am not quite sure where things ended up. looking at `BSONEncoder`, we don't actually support using a traditional options type directly. 

we do let you pass in a `CodingStrategyProvider` (under the label `options`, which is maybe kind of confusing in retrospect), but we actually store options directly on `BSONEncoder` as `var`s, and allow users to configure them (as well as `userInfo`, an option that is not part of `CodingStrategyProvider`) in that way as well.

I think I actually maybe lean toward a "no public options type" approach and just allowing configuration via mutating properties, but I don't feel that strongly. if we do keep a public struct, it seems like we should _also_ support directly mutating the properties directly on the encoder, for consistency w/ `BSONEncoder` and `JSONEncoder`.

we should probably also support the `userInfo` option, which we can pass through as the `userInfo` to the `BSONEncoder` (and maybe also `JSONEncoder`?) we use.

also, we should have notion of a default mode, which I think should be relaxed, that we use when a user does not specify one.

Sources/BSON/ExtendedJSONEncoder.swift
nspektor: done

Sources/BSON/ExtendedJSONEncoder.swift
nspektor: ok I added some more details about the different types and what they are for

Sources/BSON/ExtendedJSONEncoder.swift
nspektor: done

Sources/BSON/ExtendedJSONEncoder.swift
nspektor: @patrickfreed and I discussed this a bit and this is what he thought you landed on in the design doc discussion. I am fine with either approach

patrickfreed: Yeah, I don't feel too strongly about it either way, and given the weird situation with options on `BSONEncoder` perhaps its better to take the safer path and just include no options struct here.

>we should probably also support the userInfo option, which we can pass through as the userInfo to the BSONEncoder (and maybe also JSONEncoder?) we use.

This is a really good point, and something we definitely should do. I think we only need to pass it to the encoder that interfaces with `T`, which in this case is `BSONEncoder`. We also need to do this on the decoding end, so I filed SWIFT-959 for that work.

>also, we should have notion of a default mode, which I think should be relaxed, that we use when a user does not specify one.

Yep, and the spec recommends relaxed as the default, too. Though it also suggests having separate methods for canonical or relaxed, which we could consider as well.

nspektor: So, instead of passing anything into init, we can have 2 public variables: `mode`, and `userInfo`? 

> Yep, and the spec recommends relaxed as the default, too. Though it also suggests having separate methods for canonical or relaxed, which we could consider as well.

Does this mean having 2 encode methods? like `encodeRelaxed` and `encodeCanonical`? is that not confusing to people who expect an encoder to just have a method called `encode`?

patrickfreed: >So, instead of passing anything into init, we can have 2 public variables: mode, and userInfo?

Yep

>Does this mean having 2 encode methods? like encodeRelaxed and encodeCanonical? is that not confusing to people who expect an encoder to just have a method called encode?

Yeah, that's a good point. So it's probably best if we just stick to `encode` with a mode. We should probably document on `encode` the default value for `mode` and how users can configure it.

kmahar: > So, instead of passing anything into init, we can have 2 public variables: mode, and userInfo?

SGTM

> Does this mean having 2 encode methods? like `encodeRelaxed` and `encodeCanonical`? is that not confusing to people who expect an encoder to just have a method called `encode`?

I agree, I prefer the single method

nspektor: Okay I made those changes and added some details about the mode to the documentation.

Sources/BSON/ExtendedJSONEncoder.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// Contextual user-provided information for use during encoding.
+    public var userInfo: [CodingUserInfoKey: Any] = [:]
+
+    /// Initialize an `ExtendedJSONEncoder`.
+    public init() {
+        fatalError("unimplemented")
+    }
+
+    /// Encodes an instance of the Encodable Type `T` into Data representing canonical or relaxed extended JSON.
+    /// The value of `mode` will determine which format is used. If it is not set explicitly, relaxed will be used.
```
kmahar, line 29: maybe say `self.mode` just to be clear it's a property and not supposed to be an option to this method or something

nspektor: ok

PR #32, opened 2020-07-29
SWIFT-925 Implement ExtendedJSON Encoder API

nspektor:


Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -4,7 +4,75 @@ import Nimble
...
+
+        let regexObj = BSONRegularExpression(pattern: "p", options: "i")
+        let test = Test(x: true, y: 5, z: regexObj)
+
+        let decoded = try decoder.decode(Test.self, from: data)
+        let encoded: Data = try encoder.encode(test)
+
+        expect(decoded).to(equal(test))
+        let encodedStr = String(data: encoded, encoding: .utf8)
+        expect(encodedStr).to(contain("\"x\":true"))
```
nspektor, line 53: These tests are a bit messy and look weird, but just testing string equality didn't work because the order of the elements was not guaranteed. So I had to break it down and test that each piece was present in the encoded extJSON

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -4,7 +4,75 @@ import Nimble
...
+
+    struct Foo: Codable, Equatable {
+        let foo: Bool
+
+        enum CodingKeys: String, CodingKey {
+            case foo
+        }
+
+        func encode(to encoder: Encoder) throws {
+            let barInfo = encoder.userInfo[.barInfo] as! Bool
```
patrickfreed, line 26: since we have the explicit `== true` below, I think we can drop this force cast to prevent the test runner from crashing when this fails.

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -4,7 +4,75 @@ import Nimble
...
+}
+
 open class ExtendedJSONConversionTestCase: BSONTestCase {
+    struct Test: Codable, Equatable {
+        let x: Bool
+        let y: Int32
+        let z: BSONRegularExpression
+    }
+
+    struct Foo: Codable, Equatable {
```
patrickfreed, line 18: can we move this struct definition into the test that uses it? It will help organize this file a little bit. Ditto for the `Test` struct.

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -4,7 +4,75 @@ import Nimble
...
+        let relaxedEncoded: Data = try encoder.encode(test)
+        let relaxedEncodedStr = String(data: relaxedEncoded, encoding: .utf8)
+        expect(relaxedEncodedStr).to(contain("\"x\":true"))
+        expect(relaxedEncodedStr).to(contain("\"y\":5"))
+    }
+
+    func testExtendedJSONEncodingWithUserInfo() throws {
+        let encoder = ExtendedJSONEncoder()
+        encoder.userInfo[.barInfo] = true
+        let fooBarEncoded = try encoder.encode(Foo(foo: false))
```
patrickfreed, line 69: to make this test a little more straightforward, can we just remove the `foo` parameter from the `Foo` struct, and have the `encode` method just use the userInfo for encoding?

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -4,7 +4,75 @@ import Nimble
...
+
+        let regexStr = "{\"$regularExpression\":{\"pattern\":\"p\",\"options\":\"i\"}}"
+        let extJSON = "{\"x\":true,\"y\":{\"$numberInt\":\"5\"},\"z\":\(regexStr)}"
+        let data = extJSON.data(using: .utf8)!
+
+        let regexObj = BSONRegularExpression(pattern: "p", options: "i")
+        let test = Test(x: true, y: 5, z: regexObj)
+
+        let decoded = try decoder.decode(Test.self, from: data)
+        let encoded: Data = try encoder.encode(test)
```
patrickfreed, line 49: can we re-order this test so that it does all of the encoder tests and then all of the decoder tests? As it is right now, theyre intermingled and it's a little hard to follow.

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: done

PR #33, opened 2020-08-04
SWIFT-926 Enable extended json corpus tests

nspektor:
There are some fairly hacky things in here, and a linter error on the new array equals error.


Sources/BSON/BSONBinary.swift
```diff
@@ -205,10 +205,14 @@ extension BSONBinary: BSONValue {
 
     /// Converts this `BSONBinary` to a corresponding `JSON` in canonical extendedJSON format.
     internal func toCanonicalExtendedJSON() -> JSON {
-        [
+        var paddedSubTypeStr = String(self.subtype.rawValue, radix: 16)
+        if paddedSubTypeStr.count == 1 {
```
nspektor, line 209: Are there edge cases that this would not work for? Are there going to be supTypes that are longer than length 1 but still needed "0" padding

Sources/BSON/Date+BSONValue.swift
```diff
@@ -33,7 +33,12 @@ extension Date: BSONValue {
             self = Date(msSinceEpoch: int)
         case let .string(s):
             // relaxed extended JSON
-            guard let date = ExtendedJSONDecoder.extJSONDateFormatter.date(from: s) else {
+            // If fractional seconds are omitted in the input (length is 20 instead of 23),
+            // formatter should only account for seconds, otherwise formatter should take milliseconds into account          // Otherwise, fractional seconds SHOULD be omitted if zero.
+            let formatter = s.count == 20
```
nspektor, line 38: This hard coded length requirement is probably not ideal, any alternate ways to go about distinguishing between seconds and milliseconds formatting?

Sources/BSON/JSON.swift
```diff
@@ -201,4 +201,54 @@ extension JSON {
     }
 }
 
-extension JSON: Equatable {}
+extension JSON: Equatable {
+    /// Compares two `JSON`s for equality. In the case of an array, order is ignored.
+    static func == (lhs: JSON, rhs: JSON) -> Bool {
+        // Linter throws the following warning:
```
nspektor, line 207: Not sure what to do about this. It probably has something to do with the fact that I basically have an O(n^2) nested for loop down at lines 234-240. Not sure how else to check array equality and ignore order.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +156,101 @@ final class BSONCorpusTests: BSONTestCase {
...
+                            XCTFail("Unable to interpret degenerate_bson as Data")
+                            return
                         }
-                        let extjson = test.canonicalExtJSON.data(using: .ascii)!
-                        let jsonResult = try JSONDecoder().decode(Decimal128CanonicalExtJSON.self, from: extjson)
-                        let decimal128CorpusString = jsonResult.d.numberDecimal
 
-                        let decimal128FromString = try BSONDecimal128(decimal128CorpusString)
-                        let decimal128FromBinary = docFromCB.d!.decimal128Value!
+                        // bson_to_canonical_extended_json(dB) = cEJ
```
patrickfreed, line 191: This section and the next one for `bson_to_relaxed_extended_json` can be deleted.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +156,101 @@ final class BSONCorpusTests: BSONTestCase {
...
-                    //     // bson_to_canonical_extended_json(dB) = cEJ
-                    //     expect(try Document(fromBSON: dBData).canonicalExtendedJSON)
-                    //         .to(cleanEqual(test.canonicalExtJSON))
-
-                    //     // bson_to_relaxed_extended_json(dB) = rEJ (if rEJ exists)
-                    //     if let rEJ = test.relaxedExtJSON {
-                    //         expect(try Document(fromBSON: dBData).extendedJSON).to(cleanEqual(rEJ))
-                    //     }
-                    // }
+                        // native_to_canonical_extended_json( bson_to_native(dB) ) = cEJ
```
patrickfreed, line 205: Can we also add a comment before this one and the next one indicating that they're not in the spec yet but may be in the future, including the ticket number DRIVERS-1355?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +156,101 @@ final class BSONCorpusTests: BSONTestCase {
...
+                    guard !shouldSkip(testFile.description, test.description) else {
+                        continue
+                    }
                     let description = "\(testFile.description)-\(test.description)"
 
                     switch BSONType(rawValue: UInt8(testFile.bsonType.dropFirst(2), radix: 16)!)! {
                     case .invalid: // "top level document" uses 0x00 for the bson type
                         _ = ()
-                    // expect(try BSONDocument(fromJSON: test.string)).to(throwError(), description: description)
+                        expect(try BSONDocument(fromExtJSON: .string(test.string), keyPath: []))
```
patrickfreed, line 248: these tests should only be using the public API (since that's what they're verifying works), not any of the internals used to implement them. So we should replace this initializer with usage of the decoder here.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +156,101 @@ final class BSONCorpusTests: BSONTestCase {
...
+                            .to(throwError(), description: description)
                     case .decimal128:
                         _ = ()
-                    // expect(BSONDecimal128(test.string)).to(beNil(), description: description)
+                        expect(try BSONDecimal128(fromExtJSON: .string(test.string), keyPath: []))
```
patrickfreed, line 252: this one should actually go straight through the `BSONDecimal128("...")` initializer rather than extjson (like the commented out section does, except with `try` and `.toNot(throwError)`).

Tests/BSONTests/CommonTestUtils.swift
```diff
@@ -0,0 +1,107 @@
...
+        let jsonEnum = try jsonDecoder.decode(JSON.self, from: jsonData)
+        return jsonEnum
+    } catch {
+        fatalError("json should be decodable to jsonEnum")
+    }
+}
+
+/// Adds a custom "cleanEqual" predicate that compares Data representing JSON with a JSON string for equality
+/// after normalizing them with the "clean" function
+public func cleanEqual(_ expectedValue: String?) -> Predicate<Data> {
```
patrickfreed, line 22: I think we can just make `expectedValue` non-nil to make the body of this a little cleaner and remove the need for a `fatalError` above.

nspektor: okay, changed this

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -98,19 +98,46 @@ final class BSONCorpusTests: BSONTestCase {
 
     // swiftlint:disable:next cyclomatic_complexity
     func testBSONCorpus() throws {
+        let SKIPPED_CORPUS_TESTS = [
+            "Decimal128":
+                [
+                    "[dqbsr531] negatives (Rounded)",
+                    "[dqbsr531] negatives (Rounded)",
```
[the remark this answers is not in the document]
nspektor: removed

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -98,19 +98,46 @@ final class BSONCorpusTests: BSONTestCase {
 
     // swiftlint:disable:next cyclomatic_complexity
     func testBSONCorpus() throws {
+        let SKIPPED_CORPUS_TESTS = [
+            "Decimal128":
+                [
+                    "[dqbsr531] negatives (Rounded)",
```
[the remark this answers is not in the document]
nspektor: done

Tests/BSONTests/BSONCorpusTests.swift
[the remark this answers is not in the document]
nspektor: removed

Sources/BSON/BSONBinary.swift
nspektor: we don't need radix anymore?

Sources/BSON/Date+BSONValue.swift
nspektor: ok done

Tests/BSONTests/CommonTestUtils.swift
nspektor: okay

Sources/BSON/BSONBinary.swift
patrickfreed: Nope, the "x" in the format takes care of that.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -98,19 +98,46 @@ final class BSONCorpusTests: BSONTestCase {
...
+                    "Exact rounding"
+                ],
+            "Multiple types within the same document": ["All BSON types"],
+            "Array":
+                [
+                    "Multi Element Array with duplicate indexes",
+                    "Single Element Array with index set incorrectly to empty string",
+                    "Single Element Array with index set incorrectly to ab"
+                ],
+            "Regular Expression type": ["flags not alphabetized"]
```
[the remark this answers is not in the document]
patrickfreed: Oh, I see the problem now: that expectation should be comparing against `dBData`, not `cBData`.

Tests/BSONTests/BSONCorpusTests.swift
patrickfreed: Oh wait, sorry these are parse error tests, so this should `.to(throwError)`.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -98,19 +98,49 @@ final class BSONCorpusTests: BSONTestCase {
...
+                    "Exact rounding",
+                    "[dqbsr531] negatives (Rounded)",
+                    "[dqbsr431] check rounding modes heeded (Rounded)",
+                    "OK2",
+                    // TODO: SWIFT-965
+                    "[decq438] clamped zeros... (Clamped)",
+                    "[decq418] clamped zeros... (Clamped)"
+                ],
+            "Array":
+                [
```
patrickfreed, line 114: can we add a similar TODO for these? (ticket is SWIFT-963)

Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -370,10 +363,10 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
             .to(equal(date))
         expect(date.toCanonicalExtendedJSON()).to(equal(["$date": ["$numberLong": "500004"]]))
         // Relaxed Success case
-        let date2 = Date(msSinceEpoch: 1_356_351_330_501)
```
patrickfreed, line 373: could we add a new case instead of removing the old one here?

Tests/BSONTests/BSONCorpusTests.swift
nspektor: yep

Tests/BSONTests/BSONCorpusTests.swift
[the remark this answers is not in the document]
nspektor: but doesnt the spec say it should be compared to cB?

Sources/BSON/BSONBinary.swift
nspektor: cool

Tests/BSONTests/BSONCorpusTests.swift
nspektor: ok

Tests/BSONTests/ExtendedJSONConversionTests.swift
nspektor: sure

Tests/BSONTests/BSONCorpusTests.swift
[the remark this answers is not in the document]
patrickfreed: Hmm, I guess the spec is assuming that our native type will actually clean up the degenerate BSON, which we do not do. I think we should actually just skip that assertion altogether but also leave a comment indicating why (e.g. we just validate the BSON bytes, we don't clean them up). The actual spirit of the assertion will be implemented in SWIFT-964.

Tests/BSONTests/BSONCorpusTests.swift
patrickfreed: Yeah, that definitely shouldn't be happening. It looks like they're actually being skipped right now by the `continue` a few lines up, hence no failures.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +160,96 @@ final class BSONCorpusTests: BSONTestCase {
...
 
             if let parseErrorTests = testFile.parseErrors {
                 continue // TODO: EXT JSON support required
                 for test in parseErrorTests {
+                    guard !shouldSkip(testFile.description, test.description) else {
+                        continue
+                    }
                     let description = "\(testFile.description)-\(test.description)"
-
+                    guard let testData = Data(hexString: test.string) else {
```
patrickfreed, line 240: this `testData` portion should be moved into the "top level document" case, since for others (e.g. decimal128) `test.string` doesn't need to be converted to `Data`. Also, instead of using this `hexString` initializer, we need to be using the regular `.data(using: .utf8)`, since the string isn't a hex string but is extended JSON.

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +160,96 @@ final class BSONCorpusTests: BSONTestCase {
...
+                        XCTFail("Unable to interpret canonical_bson as Data")
+                        return
+                    }
                     switch BSONType(rawValue: UInt8(testFile.bsonType.dropFirst(2), radix: 16)!)! {
                     case .invalid: // "top level document" uses 0x00 for the bson type
                         _ = ()
```
patrickfreed, line 246: can we remove the `_ = ()` no-op here and below.

Tests/BSONTests/BSONCorpusTests.swift
nspektor: okay

Tests/BSONTests/BSONCorpusTests.swift
nspektor: yup

Tests/BSONTests/BSONCorpusTests.swift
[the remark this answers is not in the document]
nspektor: okay cool I added some comments about that but deleted the `expect()`

Tests/BSONTests/BSONCorpusTests.swift
nspektor: removed the `continue`

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +163,93 @@ final class BSONCorpusTests: BSONTestCase {
...
+                    if let rEJ = test.relaxedExtJSON {
+                        expect(try relaxedEncoder.encode(docFromCB))
+                            .to(cleanEqual(rEJ), description: test.description)
+                    }
 
-                        let decimal128FromString = try BSONDecimal128(decimal128CorpusString)
-                        let decimal128FromBinary = docFromCB.d!.decimal128Value!
+                    // for cEJ input:
+                    // native_to_canonical_extended_json( json_to_native(cEJ) ) = cEJ
+                    expect(try canonicalEncoder.encode(try decoder.decode(BSON.self, from: cEJData)))
```
kmahar, line 181: why BSON instead of BSONDocument here? I would think the data is always a document and not other BSON types?

Tests/BSONTests/BSONCorpusTests.swift
```diff
@@ -129,97 +163,93 @@ final class BSONCorpusTests: BSONTestCase {
...
-
                     switch BSONType(rawValue: UInt8(testFile.bsonType.dropFirst(2), radix: 16)!)! {
                     case .invalid: // "top level document" uses 0x00 for the bson type
-                        _ = ()
-                    // expect(try BSONDocument(fromJSON: test.string)).to(throwError(), description: description)
+                        guard let testData = test.string.data(using: .utf8) else {
+                            XCTFail("Unable to interpret canonical_bson as Data")
+                            return
+                        }
+                        expect(try decoder.decode(BSON.self, from: testData))
```
kmahar, line 249: similar q about why BSON over BSONDocument. also we could add the expected error type, I think? decoding error?

Sources/BSON/Double+BSONValue.swift
```diff
@@ -41,14 +41,31 @@ extension Double: BSONValue {
...
+    /// Helper function to make sure ExtendedJSON formatting matches Corpus Tests
+    internal func formatForExtendedJSON() -> String {
+        if self.isNaN {
+            return "NaN"
+        } else if self == Double.infinity {
+            return "Infinity"
+        } else if self == -Double.infinity {
+            return "-Infinity"
+        } else {
+            return String(describing: self).uppercased()
```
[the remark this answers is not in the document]
nspektor: Doubles with exponents in them have lower case `e` but the spec wants uppercase `E`

Tests/BSONTests/BSONCorpusTests.swift
nspektor: yep I added the error type

Tests/BSONTests/BSONCorpusTests.swift
nspektor: Yeah BSONDocument makes sense here, and probably in all of the places where I currently have `BSON.self`?

kmahar: yeah I believe so, sgtm

PR #35, opened 2020-08-06
SWIFT-973 Fix crash when encoding top level arrays to BSON

nspektor:


kmahar:
evg failed, looks like it maybe needs a rebase on top of your last commit? changes here LGTM though

nspektor:
Fixed the evergreen failures

PR #36, opened 2020-08-07
SWIFT-930 Implement existing ExtendedJSON API

nspektor:


Sources/BSON/BSONDecoder.swift
```diff
@@ -177,38 +177,26 @@ public class BSONDecoder {
...
+    public func decode<T: Decodable>(_: T.Type, from json: String) throws -> T {
+        let decoder = ExtendedJSONDecoder()
+        guard let jsonData = json.data(using: .utf8) else {
+            throw DecodingError.dataCorrupted(
+                DecodingError.Context(
+                    codingPath: [],
+                    debugDescription: "Unable to parse JSON string \(json)"
+                ))
+        }
+        return try self.decode(T.self, fromBSON: try decoder.decode(BSON.self, from: jsonData))
```
patrickfreed, line 198: rather than routing through this decoder, we can decode directly to `T` via `ExtendedJSONDecoder`:
```swift
return try decoder.decode(T.self, from: jsonData)
```

nspektor: done :)

PR #38, opened 2020-08-17
SWIFT-960 Support outputFormatting options in ExtendedJSONEncoder

patrickfreed:
SWIFT-960

This PR adds support for output formatting options on `ExtendedJSONEncoder` similar to [the ones offered on `JSONEncoder`](https://developer.apple.com/documentation/foundation/jsonencoder/outputformatting).

### DIFF
--- a/Tests/BSONTests/ExtendedJSONConversionTests.swift
+++ b/Tests/BSONTests/ExtendedJSONConversionTests.swift
@@ -107,11 +107,11 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
         let encoder = ExtendedJSONEncoder()
         let input: BSONDocument = ["topLevel": ["hello": "world"]]
 
-        let defaultFormat = String(data: try encoder.encode(input), encoding: .utf8) ?? ""
+        let defaultFormat = String(data: try encoder.encode(input), encoding: .utf8)
         expect(defaultFormat).to(equal("{\"topLevel\":{\"hello\":\"world\"}}"))
 
         encoder.outputFormatting = [.prettyPrinted]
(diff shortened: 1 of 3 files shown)


Tests/BSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -103,6 +103,41 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
         expect(String(data: fooEncoded, encoding: .utf8)).to(contain("false"))
     }
 
+    func testOutputFormatting() throws {
+        let encoder = ExtendedJSONEncoder()
+        let input: BSONDocument = ["topLevel": ["hello": "world"]]
+
+        let defaultFormat = String(data: try encoder.encode(input), encoding: .utf8) ?? ""
```
kmahar, line 110: nit: I don't think any of the `?? ""` are necessary, Nimble will let us pass in optionals and compare them to the expected non-optional strings

patrickfreed: fixed

patrickfreed:
I went to check if libbson's output was pretty printed to see if our `CustomStringConvertible` conformance matched, and I realized that we forgot to add it entirely, so I added that here. libbson's output is pretty-printed, so I enabled that in the `to(Canonical)ExtendedJSONString` methods, but the format is still slightly different. Not sure if that's a problem or not though, since the JSON itself is the same.

old:
```
{ "x" : 1 }
```
new:
```
{
  "x" : 1
}
```

kmahar:
I'm not too concerned about it. this seems similar to our conversation about key ordering. I don't think users should rely on exact string outputs always being the same over time; we only guarantee that the data will be there and will correctly represented

PR #41, opened 2020-08-25
SWIFT-914 Add docs generation script

kmahar:
This adds a new script for generating docs via Jazzy. 

As part of this I brought over most of the driver's existing BSON guide. I am thinking we can then cut out most of the driver's guide in the ticket about updating the driver docs, leaving a link to the BSON version of the guide, along with the parts of that guide that are specific to the driver (mainly about using Codable types with collections.)

### DIFF
--- a/etc/docs-main.md
+++ b/etc/docs-main.md
@@ -1,0 +1,7 @@
+# swift-bson Documentation
+
+This is the documentation for the official MongoDB Swift BSON library, [swift-bson](https://github.com/mongodb/swift-bson).
+
+You can view the README for this project, including installation instructions, [here](https://github.com/mongodb/swift-bson/blob/master/README.md).
+
+The documentation for the official MongoDB Swift driver, which depends on this BSON library, can be found [here](https://mongodb.github.io/mongo-swift-driver/MongoSwift/index.html).
--- a/etc/generate-docs.sh
+++ b/etc/generate-docs.sh
@@ -11,6 +11,9 @@ fi
 
 version=${1}
 
+# Ensure version is non-empty
+[ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
+
 jazzy_args=(--clean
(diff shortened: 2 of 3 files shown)


etc/docs-main.md
```diff
@@ -0,0 +1,7 @@
+# swift-bson Documentation
```
kmahar, line 1: I realized that, since we are generating the docs on master, this has to live on master too, oops

etc/generate-docs.sh
```diff
@@ -0,0 +1,40 @@
...
+            --documentation "Guides/*.md" 
+            --github-file-prefix https://github.com/mongodb/swift-bson/tree/v${version} 
+            --module-version "${version}" 
+            --swift-build-tool spm)
+
+modules=( BSON )
+
+for module in "${modules[@]}"; do
+  args=("${jazzy_args[@]}"  --output "docs/${module}" --module "${module}" 
+        --root-url "https://mongodb.github.io/swift-bson/docs/${module}/")
```
patrickfreed, line 29: How do we actually publish the docs to this URL?

etc/generate-docs.sh
```diff
@@ -0,0 +1,40 @@
...
+# usage: ./etc/generate-docs.sh [new version string]
+
+# exit if any command fails
+set -e
+
+if ! command -v jazzy > /dev/null; then
+  gem install jazzy || { echo "ERROR: Failed to locate or install jazzy; please install yourself with 'gem install jazzy' (you may need to use sudo)"; exit 1; }
+fi
+
+version=${1}
```
patrickfreed, line 12: Actually, can we ensure this version is non-empty? I'm worried the commit might happen without a version string below.

etc/generate-docs.sh
kmahar: the repository settings are configured such that GitHub Pages will automatically deploy the contents of the `/docs` folder on the `gh-pages` branch:
image

etc/generate-docs.sh
kmahar: good point, I added a check for this.

PR #42, opened 2020-08-25
SWIFT-937 Add a release script

kmahar:
Adds a release script. I haven't run it to completion for obvious reasons but it borrows strongly from the driver's script so I think it will work fine.

In testing this I did accidentally prematurely publish the 3.0 docs (forgot to comment out stuff in that script to commit and push...) but I think that's fine, we aren't linking to it anywhere yet.

### DIFF
--- a/etc/release.sh
+++ b/etc/release.sh
@@ -6,14 +6,18 @@
 set -e
 
 # ensure we are on master before releasing
-git checkout master
+#git checkout master
+
+version=${1}
+# Ensure version is non-empty
+[ ! -z "${version}" ] || { echo "ERROR: Missing version string"; exit 1; }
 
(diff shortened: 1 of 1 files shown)


etc/release.sh
```diff
@@ -0,0 +1,19 @@
...
+# usage: ./etc/release.sh [new version string]
+
+# exit if any command fails
+set -e
+
+# ensure we are on master before releasing
+git checkout master
+
+# regenerate documentation with new version string
+./etc/generate-docs.sh ${1}
```
patrickfreed, line 12: can we add a check to ensure a version string is passed in?

kmahar: added 👍

PR #44, opened 2020-08-27
SWIFT-995 Rename BSON target to SwiftBSON

patrickfreed:
SWIFT-995

This PR renames the `BSON` target/module to `SwiftBSON` to prevent collision with the `BSON` enum.


kmahar:
Looks like you need to fix the import sort orders (I think swiftlint autocorrect will do that). Also we should update the README to show the correct module name to depend on in the Package.swift example

PR #46, opened 2020-09-11
SWIFT-854 Add custom docs sections

kmahar:
Adds the necessary configuration info to generate the docs with custom categories.


Sources/SwiftBSON/BSONDocumentIterator.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 import NIO
 
+/// :nodoc:
```
kmahar, line 4: anything not listed in a custom category automatically gets put into an "other" section, so this was necessary to keep this type from showing up there.

etc/generate-docs.sh
```diff
@@ -14,24 +14,11 @@ version=${1}
...
-            --github_url https://github.com/mongodb/swift-bson 
-            --theme fullwidth 
-            --documentation "Guides/*.md" 
-            --github-file-prefix https://github.com/mongodb/swift-bson/tree/v${version} 
-            --module-version "${version}" 
-            --swift-build-tool spm)
-
-modules=( BSON )
-
-for module in "${modules[@]}"; do
```
kmahar, line 30: I copied this from the driver where we have multiple modules, but it's not really necessary here. I moved all of the config options into the yaml file excluding those that are dynamically determined based on the version string (and also the output directory since the use of that is rather specific to how this script works.)

.jazzy.yml
```diff
@@ -0,0 +1,47 @@
...
+theme: fullwidth
+swift_build_tool: spm
+readme: "etc/docs-main.md"
+root_url: https://mongodb.github.io/swift-bson/docs/SwiftBSON/
+documentation: "Guides/*.md"
+
+custom_categories:
+  - name: General Guides
+    children:
+      - BSON-Guide
```
kmahar, line 14: documentation files are lumped in with types, so I needed to rename this file to disambiguate it from the BSON enum.

patrickfreed (approved):

PR #48, opened 2020-10-07
SWIFT-1016 Add a public withID() method

kmahar:
This adds a `withID()` method to `BSONDocument` and does a bit of internal refactoring.

Runtime comparisons for relevant benchmarks:
| Benchmark                        | Old   | With SWIFT-1015 | With this fix |
|----------------------------------|-------|-----------------|---------------|
| Small doc bulk insert 10k copies | .096  | 0.424           | 0.116         |
| Large doc bulk insert 10 copies  | 0.332 | 2.79            | 0.352         |
| small doc insertOne 10k times    | 3.258 | 4.481           | 4.021         |
| large doc insertOne 10 times     | 0.327 | 2.812           | 0.326         |

### DIFF
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -16,21 +16,6 @@ public struct BSONDocument {
 
     private var storage: BSONDocumentStorage
 
-    internal var byteLength: Int {
-        get {
-            guard let byteLength = self.storage.buffer.getInteger(at: 0, endianness: .little, as: Int32.self) else {
-                fatalError("Cannot read byteLength of BSON from buffer")
-            }
-            return Int(byteLength)
-        }
-        set {
-            guard newValue <= Int32.max else {
-                fatalError("Cannot cast \(newValue) down to Int32")
-            }
-            self.storage.buffer.setInteger(Int32(newValue), at: 0, endianness: .little, as: Int32.self)
-        }
-    }
-
     /// An unordered set containing the keys in this document.
     internal private(set) var keySet: Set<String>
 
@@ -66,8 +51,8 @@ public struct BSONDocument {
             fatalError("Data is \(self.storage.buffer.writerIndex - start) bytes, "
                 + "but maximum allowed BSON document size is \(Int32.max) bytes")
         }
-        // Set byteLength in reserved space
-        self.storage.buffer.setInteger(byteLength, at: 0, endianness: .little, as: Int32.self)
+        // Set encodedLength in reserved space
+        self.storage.encodedLength = Int(byteLength)
     }
 
     /// Initializes a new, empty `BSONDocument`.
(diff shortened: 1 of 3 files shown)


Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -16,21 +16,6 @@ public struct BSONDocument {
 
     private var storage: BSONDocumentStorage
 
-    internal var byteLength: Int {
```
kmahar, line 19: this seemed better suited to living on the `BSONDocumentStorage` type rather than on `BSONDocument` so I've moved it there.
I also renamed it `encodedLength` as I was having a hard time keeping track of what `byteLength` meant and I think this better conveys its meaning

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -103,16 +88,13 @@ public struct BSONDocument {
...
         let storage = BSONDocumentStorage(bson)
-        try storage.validate()
-        self = BSONDocument(fromUnsafeBSON: storage)
+        let keys = try storage.validateAndRetrieveKeys()
+        self = BSONDocument(fromUnsafeBSON: storage, keys: keys)
     }
 
-    private init(fromUnsafeBSON storage: BSONDocumentStorage) {
-        self.keySet = Set()
+    private init(fromUnsafeBSON storage: BSONDocumentStorage, keys: Set<String>) {
```
kmahar, line 95: we have always already inspected the keys by the time we call this method, so we can save ourselves iterating through again by passing them in here

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -323,36 +362,44 @@ public struct BSONDocument {
             return self.buffer.writerIndex - writer
         }
 
-        internal func validate() throws {
+        @discardableResult
+        internal func validateAndRetrieveKeys() throws -> Set<String> {
```
kmahar, line 366: in order to support passing in the keys to the initializer above, I started returning the keys here.

this also led me to realize that we were not validating here that the BSON didn't contain duplicate keys so I started doing so.

Tests/SwiftBSONTests/DocumentTests.swift
```diff
@@ -758,4 +758,28 @@ final class DocumentTests: BSONTestCase {
             expect(try BSONDocument(fromBSON: data)).to(throwError(errorType: BSONError.InvalidArgumentError.self))
         }
     }
+
+    func testWithID() throws {
```
patrickfreed, line 762: can we add some assertions to this test verifying that the ObjectID was added to the front of the document?

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -240,6 +222,44 @@ public struct BSONDocument {
...
     }
 
+    /**
+     * Returns a copy of this document with an `_id` element prepended. If the document already contains an `_id`,
+     * returns the document as-is.
+     * - Throws: `BSONError.DocumentTooLargeError` if adding the `_id` would make the document exceed the maximum
+     *           allowed size for a document.
+     * - SeeAlso: https://docs.mongodb.com/manual/core/document/#the-id-field
+     */
+    public func withID() throws -> BSONDocument {
```
patrickfreed, line 232: Where are the other places that we throw for the document becoming too large? I wonder if it's rare enough that we ought to just fatalError like we do in the subscripts. Given that most users are using BSON with the driver / db, the limit in practice is what the server can take (16MB), so the ~2GB `MAX_BSON_SIZE` is never really hit.

kmahar: the only other place we throw this error is in `set(key:to:)`, which is the underlying implementation for the subscript setter method. so it ultimately always gets propagated as a `fatalError` via that path.
I guess it does seem inconsistent that this would throw but subscript doesn't, but OTOH we are actually able to propagate the error the user in the place we call this from in the driver so it is probably beneficial to do so.

patrickfreed (approved):

PR #50, opened 2020-12-01
SWIFT-1004 Support parsing $uuid as extended JSON representation for subtype 4 binary

ncarbon:


Sources/SwiftBSON/BSON.swift
```diff
@@ -95,6 +95,26 @@ public enum BSON {
             self = int64.bson
             return
         }
+
+        if let uuidJSON = try json.unwrapObject(withKey: "$uuid", keyPath: keyPath) {
```
ncarbon, line 99: I notice that the other types have the similar parsing logic abstracted, should we do something similar with `uuid`?

patrickfreed: The integers are done separately here to ensure they come first (see the comment above), but the rest of the parsing should be done via each type's conformance to `BSONValue`. This binary parsing logic can just go with the rest of the extended JSON parsing logic in in `BSONBinary.init(fromExtendedJSON:)` in `BSONBinary.swift`.

Sources/SwiftBSON/BSON.swift
```diff
@@ -95,6 +95,26 @@ public enum BSON {
...
+            guard let uuidString = uuidJSON.stringValue else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `uuid` from \"\(uuidJSON)\", "
```
patrickfreed, line 103: Can we make this description a little more specific and indicate that we were expecting a string but got some other value.

Sources/SwiftBSON/BSON.swift
```diff
@@ -95,6 +95,26 @@ public enum BSON {
...
+                )
+            }
+            guard let uuid = UUID(uuidString: uuidString) else {
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Invalid UUID string"
```
patrickfreed, line 109: can we include the invalid uuid string here? That way users can inspect the error to see what's wrong with the input string.

Sources/SwiftBSON/BSON.swift
```diff
@@ -95,6 +95,26 @@ public enum BSON {
...
+                )
+            }
+
+            let binary = try BSONBinary(from: uuid)
```
patrickfreed, line 113: Because this path is only used in the decoder, we need to repackage any errors that are thrown into `DecodingError`s. We can do that here by wrapping this line in a `do { ... } catch { ... }` that preserves the original error's description. See [here](https://github.com/mongodb/swift-bson/blob/master/Sources/SwiftBSON/BSONBinary.swift#L191) for an example of this.

Sources/SwiftBSON/BSON.swift
```diff
@@ -128,137 +148,137 @@ public enum BSON {
 }
 
 /// Value getters
-extension BSON {
+public extension BSON {
```
patrickfreed, line 151: I just merged the formatting changes, so you can revert the changes to this file and rebase and the linting tasks should pass.

patrickfreed (commented):
parsing logic looks good, just have some comments on errors and moving things around

patrickfreed (approved):
LGTM! (pending green CI)

kmahar (approved):

PR #51, opened 2020-12-03
SWIFT-344 Add equalsIgnoreOrder

ncarbon:
Adds a method for checking document equality that does not take ordering into account


Tests/SwiftBSONTests/DocumentTests.swift
```diff
@@ -237,6 +237,96 @@ final class DocumentTests: BSONTestCase {
...
+            "array2": ["string1", "string2"],
+            "null": .null,
+            "code": .code(BSONCode(code: "console.log('hi');")),
+            "maxkey": .maxKey
+        ]
+
+        // comparing two documents with the same key-value pairs in different order should return true
+        expect(a.equalsIgnoreOrder(b)).to(equal(true))
+
+        let c: BSONDocument = [
```
ncarbon, line 302: Am I over testing or is this fine? 😅

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
 extension BSONDocument: CustomStringConvertible {
     public var description: String { self.toExtendedJSONString() }
 }
+
+extension BSONDocument {
+    /// Helper function to check for BSON document equality regardless of the order of key-value pairs
+    internal func equalsIgnoreOrderHelper(a: BSONDocument, b: BSONDocument) -> Bool {
```
patrickfreed, line 523: I don't think this separate helper is necessary, it seems like we could just move the body of this function to the body of `BSONDocument.equalsIgnoreOrder`.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
+        let keysA = a.keys.sorted()
+        let keysB = b.keys.sorted()
+
+        guard keysA == keysB else {
+            return false
+        }
+
+        for k in keysA {
```
patrickfreed, line 531: rather than iterating over the `keysA` list, it would be more performant to iterate over `a` itself, since the `BSONDocumentIterator` keeps track of where it is in the document and does not need to walk over it from the beginning, like using the subscript (i.e. `a[x]`) does.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
+            guard a[k] == b[k] else {
+                return false
+            }
+        }
+
+        return true
+    }
+
+    /**
+     * Compares this `BSONDocument` with another provided one and returns true if they have the same key/value pairs
```
patrickfreed, line 547: "Compares this `BSONDocument` with another" and "returns true if they have" are a little redundant; I think we can just drop them without losing any clarity: "Returns whether this `BSONDocument` contains exactly the same key/value pairs as the provided `BSONDocument`, regardless of the order of the keys". 

I think we should also include a warning that this method is much less efficient than checking for regular equality since the document is internally ordered.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
     public var description: String { self.toExtendedJSONString() }
 }
+
+extension BSONDocument {
+    /// Helper function to check for BSON document equality regardless of the order of key-value pairs
+    internal func equalsIgnoreOrderHelper(a: BSONDocument, b: BSONDocument) -> Bool {
+        let keysA = a.keys.sorted()
+        let keysB = b.keys.sorted()
+
+        guard keysA == keysB else {
```
patrickfreed, line 527: Rather than sorting and checking for equality (which is expensive), we can simply check to see if their lengths are the same. The next portion will ensure that the keys and values are all equal in one step.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
+    /**
+     * Compares this `BSONDocument` with another provided one and returns true if they have the same key/value pairs
+     * in them regardless of the order.
+     *
+     * - Parameters:
+     *   - other: a `BSONDocument` to compare this document with.
+     *
+     * - Returns: a `Bool` indicating whether the two documents are equal.
+     */
+    public func equalsIgnoreOrder(_ other: BSONDocument) -> Bool {
```
patrickfreed, line 555: I think we may need to specify "KeyOrder", since we do want BSON arrays to be compared in the proper order.

Sources/SwiftBSON/BSONDocument.swift
ncarbon: Oh nice, I didn't realize that 😅

Sources/SwiftBSON/BSONDocument.swift
ncarbon: Oops, yeah that was unnecessary. Fixed 👍🏼

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
+     *
+     * - Parameters:
+     *   - other: a `BSONDocument` to compare this document with.
+     *
+     * - Returns: a `Bool` indicating whether the two documents are equal.
+     */
+    public func equalsIgnoreKeyOrder(_ other: BSONDocument) -> Bool {
+        let a = self
+        let b = other
+        let keysA = a.keys.sorted()
```
kmahar, line 537: I don't think we need to sort these before checking if the counts are equal.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,42 @@ extension BSONDocument: BSONValue {
...
+        }
+
+        for (k, v) in self {
+            if case let (.document(aDoc), .document(bDoc)?) = (v, other[k]) {
+                guard aDoc.equalsIgnoreKeyOrder(bDoc) else {
+                    return false
+                }
+                continue
+            }
+            guard v == b[k] else {
```
kmahar, line 551: we seem to be mixing usage of `other` and `b` here. we can probably get rid of `a` and `b` above and just use `self` and `other` directly

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -517,3 +517,37 @@ extension BSONDocument: BSONValue {
 extension BSONDocument: CustomStringConvertible {
     public var description: String { self.toExtendedJSONString() }
 }
+
+extension BSONDocument {
+    /**
+     * Returns whether this BSONDocument contains exactly the same key/value pairs as the provided BSONDocument,
```
kmahar, line 523: nit

```suggestion
     * Returns whether this `BSONDocument` contains exactly the same key/value pairs as the provided `BSONDocument`,
```

kmahar (approved):
one nit LGTM otherwise!

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -519,40 +519,35 @@ extension BSONDocument: CustomStringConvertible {
...
+     * Warning: This method is much less efficient than checking for regular equality since the document is internally
+     * ordered.
+     *
+     * - Parameters:
+     *   - other: a `BSONDocument` to compare this document with.
+     *
+     * - Returns: a `Bool` indicating whether the two documents are equal.
+     */
+    public func equalsIgnoreKeyOrder(_ other: BSONDocument) -> Bool {
+        guard self.keys.count == other.keys.count else {
```
patrickfreed, line 535: `BSONDocument.keys` returns a `[String]` containing all the keys in the document in order, meaning it requires a traversal of the entire document (i.e. O(n)). `BSONDocument.count` instead uses the cached `keySet`, so no traversal is required (i.e. O(1)).

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -519,40 +519,35 @@ extension BSONDocument: CustomStringConvertible {
...
             return false
         }
 
-        for k in keysA {
-            if case let (.document(aDoc), .document(bDoc)?) = (a[k], b[k]) {
-                guard self.equalsIgnoreOrderHelper(a: aDoc, b: bDoc) else {
+        for (k, v) in self {
+            if case let (.document(docA), .document(docB)?) = (v, other[k]) {
```
patrickfreed, line 540: to save on traversing the document multiple times, we should compute `other[v]` once and then use it in the various comparisons.

patrickfreed (commented):
Looks good, just a few more perf suggestions

patrickfreed (approved):
LGTM!

PR #52, opened 2021-01-08
SWIFT-1026 Extended JSON parsing performance improvements

patrickfreed:
SWIFT-1026

This PR implements the proposed improvements to Extended JSON parsing in `swift-bson` as part of SWIFT-993.

The main improvements are as follows:
- replaced Foundation's `JSONDecoder` with the much faster `XJSONDecoder` from [`swift-extras-json`](https://github.com/swift-extras/swift-extras-json)
- decode specific BSON types based on the input JSON rather trying each successively until one works
- Recursively build up a single document reusing the same ByteBuffer rather than separate ones for each subdocument
- replace Foundation's base64 decoding with the much faster [`swift-extras-base64`](https://github.com/swift-extras/swift-extras-base64)

Re-running the benchmarks on the spawn host, both the flat BSON and deep BSON ones are within the target time, though the full BSON one exceeds it by a little bit due to code and codewithscope parsing being slow. I figured it probably wasn't worth putting in much effort to improve this, however.

I reran the baseline again to get updated numbers. 

Benchmarking results (all in seconds):
| benchmark | libbson based median time | target time (4x) | post-optimizations |
|-----------|---------------------------|------------------|--------------------|
| flat JSON | 0.793                     | 3.172            | 2.513              |
| deep JSON | 0.3                       | 1.2              | 1.159              |
| full JSON | 0.789                     | 3.156            | 3.756              |

### DIFF
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -370,19 +370,25 @@ public struct BSONDocument {
             return self.buffer.writerIndex - writer
         }
 
-        /// Build a document at the current position in the storage via the provided closure which returns
-        /// how many bytes it wrote.
+        /// Build a document at the current position in the storage via the provided closure which appends
+        /// the elements of the document and returns how many bytes it wrote in total. This method will append the
+        /// required metadata surrounding the document as necessary (length, null byte).
+        ///
+        /// If this method is used to build a subdocument, the caller is responsible for updating
+        /// the length of the containing document based on this method's return value. If this method was invoked
+        /// recursively from `buildDocument`, such updating will happen automatically if the returned byte count
+        /// is propagated.
         ///
         /// This may be used to build up a fresh document or a subdocument.
-        internal mutating func buildDocument(_ appendFunc: (inout Self) throws -> Int) rethrows -> Int {
+        internal mutating func buildDocument(_ appendElementsFunc: (inout Self) throws -> Int) rethrows -> Int {
             var totalBytes = 0
 
-            // write length of document
+            // write placeholder length of document
             let lengthIndex = self.buffer.writerIndex
             totalBytes += self.buffer.writeInteger(0, endianness: .little, as: Int32.self)
 
             // write contents
-            totalBytes += try appendFunc(&self)
+            totalBytes += try appendElementsFunc(&self)
 
             // write null byte
             totalBytes += self.buffer.writeInteger(0, as: UInt8.self)
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -1,3 +1,4 @@
+import ExtrasJSON
 import Foundation
 import NIO
 
@@ -92,7 +93,7 @@ public struct BSONDocument {
         self = BSONDocument(fromUnsafeBSON: storage, keys: keys)
     }
 
-    private init(fromUnsafeBSON storage: BSONDocumentStorage, keys: Set<String>) {
+    internal init(fromUnsafeBSON storage: BSONDocumentStorage, keys: Set<String>) {
         self.storage = storage
         self.keySet = keys
     }
@@ -329,7 +330,7 @@ public struct BSONDocument {
 
     /// Storage management for BSONDocuments.
     /// A wrapper around a ByteBuffer providing various BSONDocument-specific utilities.
-    private struct BSONDocumentStorage {
+    internal struct BSONDocumentStorage {
         internal var buffer: ByteBuffer
 
         /// Create BSONDocumentStorage from ByteBuffer.
@@ -356,10 +357,39 @@ public struct BSONDocument {
         /// Appends element to underlying BSON bytes, returns the size of the element appended: type + key + value
         @discardableResult internal mutating func append(key: String, value: BSON) -> Int {
             let writer = self.buffer.writerIndex
-            self.buffer.writeInteger(value.bsonValue.bsonType.rawValue, as: UInt8.self)
-            self.buffer.writeCString(key)
+            self.appendElementHeader(key: key, bsonType: value.bsonValue.bsonType)
             value.bsonValue.write(to: &self.buffer)
             return self.buffer.writerIndex - writer
+        }
+
+        /// Append the header (key and BSONType) for a given element.
+        @discardableResult internal mutating func appendElementHeader(key: String, bsonType: BSONType) -> Int {
+            let writer = self.buffer.writerIndex
+            self.buffer.writeInteger(bsonType.rawValue, as: UInt8.self)
+            self.buffer.writeCString(key)
+            return self.buffer.writerIndex - writer
+        }
+
+        /// Build a document at the current position in the storage via the provided closure which returns
+        /// how many bytes it wrote.
+        ///
+        /// This may be used to build up a fresh document or a subdocument.
+        internal mutating func buildDocument(_ appendFunc: (inout Self) throws -> Int) throws -> Int {
+            var totalBytes = 0
+
+            // write length of document
+            let lengthIndex = self.buffer.writerIndex
+            totalBytes += self.buffer.writeInteger(0, endianness: .little, as: Int32.self)
+
+            // write contents
+            totalBytes += try appendFunc(&self)
+
+            // write null byte
+            totalBytes += self.buffer.writeInteger(0, as: UInt8.self)
+
+            self.buffer.setInteger(Int32(totalBytes), at: lengthIndex, endianness: .little, as: Int32.self)
+
+            return totalBytes
         }
 
         @discardableResult
@@ -447,8 +477,11 @@ extension BSONDocument: Equatable {
 }
 
 extension BSONDocument: BSONValue {
+    internal static let extJSONTypeWrapperKeys: [String] = []
+
     /*
      * Initializes a `BSONDocument` from ExtendedJSON.
+     * This is not as performant as ExtendedJSONDecoder.decode, so it should only be used for small documents.
      *
      * Parameters:
      *   - `json`: a `JSON` representing the canonical or relaxed form of ExtendedJSON for any `BSONDocument`.
@@ -463,12 +496,12 @@ extension BSONDocument: BSONValue {
      */
     internal init?(fromExtJSON json: JSON, keyPath: [String]) throws {
         // canonical and relaxed extended JSON
-        guard case let .object(obj) = json else {
+        guard case let .object(obj) = json.value else {
             return nil
         }
         var doc: [(String, BSON)] = []
         for (key, val) in obj {
-            let bsonValue = try BSON(fromExtJSON: val, keyPath: keyPath + [key])
+            let bsonValue = try BSON(fromExtJSON: JSON(val), keyPath: keyPath + [key])
             doc.append((key, bsonValue))
         }
         self = BSONDocument(keyValuePairs: doc)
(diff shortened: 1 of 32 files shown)


Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -28,21 +56,138 @@ public class ExtendedJSONDecoder {
...
+        let bson = try self.decodeBSONFromJSON(json, keyPath: [])
+        // let bson = BSON(fromExtJSON: json)
 
         // The `BSON` is then passed through a `BSONDecoder` where it is outputted as a `T`
         let bsonDecoder = BSONDecoder()
         bsonDecoder.userInfo = self.userInfo
         return try bsonDecoder.decode(T.self, fromBSON: bson)
     }
+
+    /// Decode a `BSON` from the given extended JSON.
```
patrickfreed, line 75: the methods here enable us to recursively build up the document without having to allocate many `ByteBuffers` for each subdocument, which is much faster. The caveat is that it is a lot less natural to do so. For scalars, we can operate pretty similarly to before, so we return those as `BSON`. (technically we should build up code with scope similarly, but I opted to treat it as a scalar for simplicity). For non-scalars (arrays and documents), we return them from `decodeScalar` as-is and then do further processing here in `decodeBSONFromJSON` and recursive calls in descending from `appendElement`.

Sources/SwiftBSON/JSON.swift
```diff
@@ -1,55 +1,29 @@
...
-        } else if container.decodeNil() {
-            self = .null
-        } else {
-            throw DecodingError.dataCorrupted(
-                DecodingError.Context(
-                    codingPath: decoder.codingPath,
-                    debugDescription: "Not a valid JSON type"
-                ))
-        }
+internal struct JSON {
```
patrickfreed, line 4: I kept this wrapper struct around so that we could add the `ExpressibleByXLiteral` and other conformances to `JSONValue`, which we get from ExtrasJSON.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -356,12 +357,41 @@ public struct BSONDocument {
...
+            self.buffer.writeInteger(bsonType.rawValue, as: UInt8.self)
+            self.buffer.writeCString(key)
+            return self.buffer.writerIndex - writer
+        }
+
+        /// Build a document at the current position in the storage via the provided closure which returns
+        /// how many bytes it wrote.
+        ///
+        /// This may be used to build up a fresh document or a subdocument.
+        internal mutating func buildDocument(_ appendFunc: (inout Self) throws -> Int) throws -> Int {
```
patrickfreed, line 377: This method allows us build up a document or subdocument without making new `ByteBuffers`. See the additions and my comment in `ExtendedJSONDecoder` to see how this is used.

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -13,6 +15,32 @@ public class ExtendedJSONDecoder {
...
+        Set(ExtendedJSONDecoder.wrapperKeyMap.keys)
+    }()
+
+    /// A map from extended JSON wrapper keys (e.g. "$numberLong") to the BSON type(s) that they correspond to.
+    ///
+    /// Some types are associated with multiple wrapper keys (e.g. "$code" and "$scope" both map to
+    /// `BSONCodeWithScope`). Some wrapper keys are associated with multiple types (e.g. "$code" maps to both
+    /// `BSONCode` and `BSONCodeWithScope`). Attempt to decode each of the types returned from the map until one works
+    /// to find the proper decoding.
+    private static var wrapperKeyMap: [String: [BSONValue.Type]] = {
```
patrickfreed, line 29: this allows us to use the input JSON to determine which type(s) to attempt to decode rather than just trying each in succession. Code and CodeWithScope make this a little trickier, hence it not being a 1:1 mapping.

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -28,21 +56,138 @@ public class ExtendedJSONDecoder {
...
-        // In earlier versions of Swift, JSONDecoder doesn't support decoding "fragments" at the top level, so we wrap
-        // the data in an array to guarantee it always decodes properly.
-        let wrappedData = "[".utf8 + data + "]".utf8
-        let json = try JSONDecoder().decode([JSON].self, from: wrappedData)[0]
+        let json = try JSONParser().parse(bytes: data)
 
         // Then a `BSON` enum instance is created via the `JSON`.
-        let bson = try BSON(fromExtJSON: json, keyPath: [])
+        // let bson = try json.toBSON(keyPath: [])
+        let bson = try self.decodeBSONFromJSON(json, keyPath: [])
```
patrickfreed, line 66: I just realized that I forgot to include the error conversions to `DecodingError` in here, will submit a patch shortly to fix this.

patrickfreed:
So it turns out the `swift-extras-base64` package might not usable for us. The base64 included in the BSON corpus tests doesn't get round-tripped properly, which seems to indicate that package is broken. I filed an [issue](https://github.com/swift-extras/swift-extras-base64/issues/22) and once that gets fixed we can re-incorporate it to regain some of those performance improvements.

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -13,6 +15,32 @@ public class ExtendedJSONDecoder {
         return formatter
     }()
 
+    /// A set of all the possible extenededJSON wrapper keys.
```
kmahar, line 18: typo "exteneded"

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -13,6 +15,32 @@ public class ExtendedJSONDecoder {
...
+    private static var wrapperKeySet: Set<String> = {
```
kmahar, line 19: does this (and the map) only get computed one time?

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -13,6 +15,32 @@ public class ExtendedJSONDecoder {
...
+    ///
+    /// Some types are associated with multiple wrapper keys (e.g. "$code" and "$scope" both map to
+    /// `BSONCodeWithScope`). Some wrapper keys are associated with multiple types (e.g. "$code" maps to both
+    /// `BSONCode` and `BSONCodeWithScope`). Attempt to decode each of the types returned from the map until one works
+    /// to find the proper decoding.
+    private static var wrapperKeyMap: [String: [BSONValue.Type]] = {
+        var map: [String: [BSONValue.Type]] = [:]
+        for t in BSON.allBSONTypes.values {
+            for k in t.extJSONTypeWrapperKeys {
+                if var existingList = map[k] {
```
kmahar, line 33: this is a nit, but since we're focused on perf anyway here... I think this _might_ be a bit more efficient since it won't incur CoW behavior for the array(I think that would happen as written since you modify `existingList` before writing it back):
```swift
map[k, default: []].append(t.self)
```

I always thought this syntax wouldn't properly append the element in place, but I learned from doing the advent of code puzzles in Swift that it does  🤷‍♀️

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -356,12 +357,41 @@ public struct BSONDocument {
...
+            let lengthIndex = self.buffer.writerIndex
+            totalBytes += self.buffer.writeInteger(0, endianness: .little, as: Int32.self)
+
+            // write contents
+            totalBytes += try appendFunc(&self)
+
+            // write null byte
+            totalBytes += self.buffer.writeInteger(0, as: UInt8.self)
+
+            self.buffer.setInteger(Int32(totalBytes), at: lengthIndex, endianness: .little, as: Int32.self)
```
kmahar, line 390: do we need to update the overall encoded length, too? i.e. the first 4 bytes of the storage

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -28,21 +56,140 @@ public class ExtendedJSONDecoder {
...
+            }
+            return .array(bsonArr)
+        case let .encodedObject(obj):
+            var storage = BSONDocument.BSONDocumentStorage()
+            _ = try self.appendObject(obj, to: &storage, keyPath: keyPath)
+            return .document(BSONDocument(fromUnsafeBSON: storage, keys: Set(obj.keys)))
+        }
+    }
+
+    /// Decode and append the given extended JSON object to the provided BSONDocumentStorage.
```
kmahar, line 92: nit: document the return value, here and for `appendElement`?

Tests/SwiftBSONTests/JSONTests.swift
```diff
@@ -1,72 +1,59 @@
...
-    let decoder = JSONDecoder()
+    let encoder = XJSONEncoder()
+    let decoder = XJSONDecoder()
 
     func testInteger() throws {
         // Initializing a JSON with an int works, but it will be cast to a double.
         let intJSON: JSON = 12
-        let encoded = try encoder.encode([intJSON])
+        let encoded = Data(try encoder.encode([intJSON]))
         /* JSONEncoder currently cannot encode non-object/array top level values.
```
kmahar, line 16: does this limitation apply to XJSONEncoder as well?

Tests/SwiftBSONTests/JSONTests.swift
```diff
@@ -1,72 +1,59 @@
...
          To get around this, the generated JSON will need to be wrapped in an array
          and unwrapped again at the end as a workaround.
          This workaround can be removed when Swift 5.3 is the minimum supported version by the BSON library. */
         expect(Double(String(data: encoded.dropFirst().dropLast(), encoding: .utf8)!)!)
             .to(beCloseTo(12))
-
```
kmahar, line 21: why have these decoding checks been removed throughout this file?

kmahar (commented):
looks good! and based on the original runtimes I had anyway you've achieved anywhere from 70-285x speedup on all of the benchmarks which is awesome. this is mainly a bunch of nits and questions

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -356,12 +357,41 @@ public struct BSONDocument {
...
+
+        /// Build a document at the current position in the storage via the provided closure which returns
+        /// how many bytes it wrote.
+        ///
+        /// This may be used to build up a fresh document or a subdocument.
+        internal mutating func buildDocument(_ appendFunc: (inout Self) throws -> Int) rethrows -> Int {
+            var totalBytes = 0
+
+            // write length of document
+            let lengthIndex = self.buffer.writerIndex
```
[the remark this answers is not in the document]
kmahar: thanks, that is helpful! 

I think what I was imagining with my concern here and the other comment was that someone could do something like
```swift
var doc: BSONDocument = ["a": 1]
doc.storage.buildDocument { storage in
    // add some stuff
}
```
in which case I don't think we'd end up with a valid top level doc, but we aren't doing that anywhere.

kmahar (approved):
LGTM! nice work

Sources/SwiftBSON/BSONDocument.swift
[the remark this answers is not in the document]
patrickfreed: Oh I see what you mean. Yeah that would be an issue; this method is really only designed to be used during document creation. To prevent something like this at the type level, we could introduce a `BSONDocumentStorageBuilder` or something, but I imagine it'd have some overlap / duplication with the actual `BSONDocumentStorage`.

PR #54, opened 2021-01-21
SWIFT-1070 Improve performance of BSONDecoder

patrickfreed:
SWIFT-1070

This PR improves the performance of the `BSONDecoder`. The goal of this PR is simply to match the libbson-based BSON library's performance, so there are still more possible improvements to be done in the future.

We actually ended up improving upon the existing libbson-based `BSONDecoder` pretty significantly:
### BSON to Native Benchmark Results
| benchmark | libbson based median time | target time (1.25x) | post-optimizations |
|-----------|---------------------------|------------------|--------------------|
| flat BSON | 6.002                     | 7.503            | 2.072              |
| deep BSON | 3.463                     | 4.329            | 2.683              |
| full BSON | 3.876                     | 4.845            | 2.017              |

We also improved upon the improved Native to BSON benchmarks we originally saw in `swift-bson`:
### Native to BSON Benchmark Results
| benchmark | libbson based median time | target time (1.25x) | unoptimized swift-bson | post-optimizations |
|-----------|---------------------------|------------------|------------------------|--------------------|
| flat BSON | 6.793                     | 8.491            | 2.789                  | 1.456              |
| deep BSON | 4.117                     | 5.146            | 3.504                  | 2.918              |
| full BSON | 6.399                     | 7.999            | 4.479                  | 2.195              |

### DIFF
--- a/Sources/SwiftBSON/Array+BSONValue.swift
+++ b/Sources/SwiftBSON/Array+BSONValue.swift
@@ -46,7 +46,12 @@ extension Array: BSONValue where Element == BSON {
         guard let doc = try BSONDocument.read(from: &buffer).documentValue else {
             throw BSONError.InternalError(message: "BSON Array cannot be read, failed to get documentValue")
         }
-        return .array(doc.values)
+        var values: [BSON] = []
+        let it = doc.makeIterator()
+        while let (_, val) = try it.nextThrowing() {
+            values.append(val)
+        }
+        return .array(values)
     }
 
     internal func write(to buffer: inout ByteBuffer) {
@@ -55,5 +60,11 @@ extension Array: BSONValue where Element == BSON {
             array[String(index)] = value
         }
         array.write(to: &buffer)
+    }
+
+    internal func validate() throws {
+        for v in self {
+            try v.bsonValue.validate()
+        }
     }
 }
--- a/Sources/SwiftBSON/BSONDocument+Sequence.swift
+++ b/Sources/SwiftBSON/BSONDocument+Sequence.swift
@@ -15,7 +15,7 @@ extension BSONDocument: Sequence {
     public typealias SubSequence = BSONDocument
 
     /// Returns a `Bool` indicating whether the document is empty.
-    public var isEmpty: Bool { self.keySet.isEmpty }
+    public var isEmpty: Bool { self.storage.encodedLength == BSON_MIN_SIZE }
 
     /// Returns a `DocumentIterator` over the values in this `Document`.
     public func makeIterator() -> BSONDocumentIterator {
@@ -29,7 +29,11 @@ extension BSONDocument: Sequence {
     public func map<T>(
         _ transform: (Element) throws -> T
     ) rethrows -> [T] {
-        try AnySequence(self).map(transform)
+        var values: [T] = []
+        for (k, v) in self {
+            values.append(try transform((key: k, value: v)))
+        }
+        return values
     }
 
     /**
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -15,27 +15,14 @@ public struct BSONDocument {
     /// The element type of a document: a tuple containing an individual key-value pair.
     public typealias KeyValuePair = (key: String, value: BSON)
 
-    private var storage: BSONDocumentStorage
-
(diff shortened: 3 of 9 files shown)


Sources/SwiftBSON/Array+BSONValue.swift
```diff
@@ -46,7 +46,12 @@ extension Array: BSONValue where Element == BSON {
         guard let doc = try BSONDocument.read(from: &buffer).documentValue else {
             throw BSONError.InternalError(message: "BSON Array cannot be read, failed to get documentValue")
         }
-        return .array(doc.values)
+        var values: [BSON] = []
+        let it = doc.makeIterator()
+        while let (_, val) = try it.nextThrowing() {
```
patrickfreed, line 51: since `BSONDocument.read` doesn't validate anymore, we need to use `nextThrowing` to ensure we don't `fatalError`. Full validation will be performed in `Array.validate` if necessary.

Sources/SwiftBSON/BSONDocument+Sequence.swift
```diff
@@ -29,7 +29,11 @@ extension BSONDocument: Sequence {
     public func map<T>(
         _ transform: (Element) throws -> T
     ) rethrows -> [T] {
-        try AnySequence(self).map(transform)
```
patrickfreed, line 32: This was calling `count` a lot, which now requires a full walk over the document. To avoid any issues, I just did a naive manual implementation of `map`.

Sources/SwiftBSON/BSONEncoder.swift
```diff
@@ -822,47 +806,47 @@ private class MutableDictionary: BSONValue {
...
-            switch value {
-            case let val as MutableDictionary:
-                try doc.set(key: self.keys[i], to: val.toDocument().bson)
-            case let val as MutableArray:
-                let array = try val.toBSONArray()
-                try doc.set(key: self.keys[i], to: array.bson)
-            default:
-                try doc.set(key: self.keys[i], to: value.bson)
-            }
+            doc.append(key: self.keys[i], value: value.bson)
```
patrickfreed, line 849: since we manually handle overwriting keys in the keyed encoding, we can just append directly without checking prior bytes of the document. This means we don't have to `throw` in here anymore and can use the `.bson` computed property.

patrickfreed:
Oh I forgot to mention: I left off the improvements to `BSONDocumentIterator` that make use of the unsafe `ByteBuffer` functionality since they are primarily used to speed up `withID` / the insertion benchmarks. The next PR I put up for insertion benchmarks will include those changes.

kmahar (commented):
looking good! just a couple questions

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -15,28 +15,15 @@ public struct BSONDocument {
...
-    internal private(set) var keySet: Set<String>
+    internal var storage: BSONDocumentStorage
 
     internal init(_ elements: [BSON]) {
         self = BSONDocument(keyValuePairs: elements.enumerated().map { i, element in (String(i), element) })
     }
 
     internal init(keyValuePairs: [(String, BSON)]) {
-        self.keySet = Set(keyValuePairs.map { $0.0 })
-        guard self.keySet.count == keyValuePairs.count else {
```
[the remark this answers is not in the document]
patrickfreed: good catch. Since this is primarily a concern of dictionary literal initializers, I've moved the validation into them instead of here, since this initializer is also used elsewhere (on potentially un-validated documents in the future).

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -166,7 +157,9 @@ public struct BSONDocument {
     public func toData() -> Data { Data(self.storage.buffer.readableBytesView) }
 
     /// Returns a `Boolean` indicating whether this `BSONDocument` contains the provided key.
-    public func hasKey(_ key: String) -> Bool { self.keySet.contains(key) }
+    public func hasKey(_ key: String) -> Bool {
+        (try? BSONDocumentIterator.find(key: key, in: self)) != nil
```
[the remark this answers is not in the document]
patrickfreed: yep, that's coming in the next one

kmahar (approved):
lgtm! nice work, exciting to end up with better perf than we had with libbson.

PR #55, opened 2021-01-22
SWIFT-1072 Improve insertion performance

patrickfreed:
SWIFT-1072

This PR includes the BSON library portion of the insertion improvements. A subsequent PR on the driver repo will be opened after this one goes through.

| benchmark             | libbson based median time | target time (4x) | unoptimized swift-bson | post-optimizations |
|-----------------------|---------------------------|------------------|------------------------|--------------------|
| small doc bulk insert | 0.096                     | 0.12             | 4258.673               | 0.117              |
| large doc bulk insert | 0.332                     | 0.415            | 59.637                 | 0.454              |
| small doc insertOne   | 3.258                     | 4.0725           | 8.276                  | 3.412              |
| large doc insertOne   | 0.327                     | 0.409            | 28.742                 | 0.456              |

As noted in the design, we missed on the large doc benchmarks by a little, but those ones are still super fast compared to other MongoDB drivers, so it isn't a huge concern.

### DIFF
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -79,6 +79,21 @@ public struct BSONDocument {
         self = BSONDocument(fromUnsafeBSON: storage)
     }
 
+    /**
+     * Initialize a new `BSONDocument` from the provided BSON data without validating the elements. The first four
+     * bytes must accurately contain the length, however.
+     *
+     * If invalid BSON data is provided, undefined behavior or server-side errors may occur when using the
+     * resultant `BSONDocument`.
+     *
+     * - Throws: `BSONError.InvalidArgumentError` if the provided BSON's length does not match the encoded length.
+     */
+    public init(withoutValidatingElements bson: ByteBuffer) throws {
+        let storage = BSONDocumentStorage(bson)
+        try storage.validateLength()
+        self.init(fromUnsafeBSON: storage)
+    }
+
     internal init(fromUnsafeBSON storage: BSONDocumentStorage) {
         self.storage = storage
     }
@@ -129,11 +144,7 @@ public struct BSONDocument {
 
     /// The keys in this `BSONDocument`.
     public var keys: [String] {
-        do {
-            return try BSONDocumentIterator.getKeys(from: self.storage.buffer)
-        } catch {
-            fatalError("Failed to retrieve keys for document")
-        }
+        BSONDocumentIterator.getKeys(from: self.storage.buffer)
     }
 
     /// The values in this `BSONDocument`.
@@ -141,11 +152,7 @@ public struct BSONDocument {
 
     /// The number of (key, value) pairs stored at the top level of this document.
     public var count: Int {
-        do {
-            return try BSONDocumentIterator.getKeys(from: self.storage.buffer).count
-        } catch {
-            return 0
-        }
+        BSONDocumentIterator.getKeys(from: self.storage.buffer).count
     }
 
     /// A copy of the `ByteBuffer` backing this document, containing raw BSON data. As `ByteBuffer`s implement
@@ -158,7 +165,7 @@ public struct BSONDocument {
 
     /// Returns a `Boolean` indicating whether this `BSONDocument` contains the provided key.
     public func hasKey(_ key: String) -> Bool {
-        (try? BSONDocumentIterator.find(key: key, in: self)) != nil
+        BSONDocumentIterator.find(key: key, in: self) != nil
     }
 
     /**
@@ -174,19 +181,10 @@ public struct BSONDocument {
      */
     public subscript(key: String) -> BSON? {
         get {
-            do {
-                return try BSONDocumentIterator.find(key: key, in: self)?.value
-            } catch {
-                fatalError("Error looking up key \(key) in document: \(error)")
-            }
+            BSONDocumentIterator.find(key: key, in: self)?.value
         }
         set {
-            // The only time this would crash is document too big error
-            do {
-                return try self.set(key: key, to: newValue)
-            } catch {
-                fatalError("Failed to set \(key) to \(String(describing: newValue)): \(error)")
-            }
+            self.set(key: key, to: newValue)
         }
     }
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -74,12 +74,27 @@ public struct BSONDocument {
...
+    /**
+     * Initialize a new `BSONDocument` from the provided BSON data without validating the elements. The first four
+     * bytes must accurately reflect the length of the buffer, however.
+     *
+     * If invalid BSON data is provided, undefined behavior or server-side errors may occur when using the
+     * resultant `BSONDocument`.
+     *
+     * - Throws: `BSONError.InvalidArgumentError` if the provided BSON's length does not match the encoded length.
+     */
+    public init(fromBSONWithoutValidatingElements bson: ByteBuffer) throws {
```
patrickfreed, line 91: this will be the initializer that we use in the driver when dealing with BSON from libmongoc. It has to be a part of the public API since we use it in our other package, but I think that's okay because I can imagine users may find themselves in a similar situation.

I'm open to less verbose / different names for this by the way. I chose this one since I figured we might as well make it as explicit as possible if we don't want users to accidentally use this initializer wrongly.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -74,12 +74,27 @@ public struct BSONDocument {
...
+        let storage = BSONDocumentStorage(bson)
+        try self.init(fromBSONWithoutValidatingElements: storage)
     }
 
-    internal init(fromUnsafeBSON storage: BSONDocumentStorage) {
+    internal init(fromBSONWithoutValidatingElements storage: BSONDocumentStorage) throws {
+        try storage.validateLength()
```
patrickfreed, line 97: Since it's basically free to validate the length and it makes the iterator logic a lot safer, it made sense to do it as a bare minimum validation.

Sources/SwiftBSON/BSONDocumentIterator.swift
```diff
@@ -22,12 +22,10 @@ public class BSONDocumentIterator: IteratorProtocol {
```
patrickfreed, line 22: A lot of the changes in this file are for making it safe to use an iterator with invalid BSON, defaulting to just returning early or nil when errors are encountered.

Sources/SwiftBSON/BSONDocumentIterator.swift
```diff
@@ -49,12 +47,13 @@ public class BSONDocumentIterator: IteratorProtocol {
     /// Get the next key in the iterator, if there is one.
     /// This method should only be used for iterating through the keys. It advances to the beginning of the next
     /// element, meaning the element associated with the last returned key cannot be accessed via this iterator.
-    private func nextKey() throws -> String? {
-        guard let type = try self.readNextType() else {
+    private func nextKey() -> String? {
```
patrickfreed, line 50: any methods not called from `nextThrowing` (which is used in validation and thus has useful error messages) I removed `throws` from and instead just returned `nil`.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -271,8 +269,8 @@ public struct BSONDocument {
      * Sets a BSON element with the corresponding key
      * if element.value is nil the element is deleted from the BSON
      */
-    internal mutating func set(key: String, to value: BSON?) throws {
-        guard let range = try BSONDocumentIterator.findByteRange(for: key, in: self) else {
+    internal mutating func set(key: String, to value: BSON?) {
```
kmahar, line 272: I think we only call this method from the subscript setter, could we make it `private`?

Tests/SwiftBSONTests/BSONCorpusTests.swift
```diff
@@ -272,6 +272,19 @@ final class BSONCorpusTests: BSONTestCase {
                         return
                     }
                     expect(try BSONDocument(fromBSON: data)).to(throwError(), description: description)
+
```
kmahar, line 275: nice idea to use the corpus for testing the behavior 👍

Sources/SwiftBSON/BSONDocumentIterator.swift
kmahar: I think it would be helpful for future us to note throughout the docstrings in the file what the method's behaviors are when invalid BSON is encountered

Sources/SwiftBSON/BSONDocumentIterator.swift
```diff
@@ -95,95 +94,173 @@ public class BSONDocumentIterator: IteratorProtocol {
         return bsonType
     }
 
+    /// Search for the value associated with the given key, returning its type if found and nil otherwise.
+    /// This moves the iterator right up to the first byte of the value.
+    internal func findValue(forKey key: String) -> BSONType? {
```
kmahar, line 99: could we possibly match by accident on the right string but in the wrong place? like if a subdocument contains the same key, or a string value contains it, etc.

Sources/SwiftBSON/BSONDocumentIterator.swift
patrickfreed: done

Sources/SwiftBSON/BSONDocumentIterator.swift
patrickfreed: I don't think so, since we use `skipNextValue` between each attempt to read a key.

Sources/SwiftBSON/BSONDocumentIterator.swift
```diff
@@ -95,126 +96,197 @@ public class BSONDocumentIterator: IteratorProtocol {
...
         }
+
+        return true
     }
 
     /// Finds the key in the underlying buffer, and returns the [startIndex, endIndex) containing the corresponding
     /// element.
-    internal static func findByteRange(for searchKey: String, in document: BSONDocument) throws -> Range<Int>? {
+    /// Returns nil if invalid BSON is encountered.
+    internal static func findByteRange(for searchKey: String, in document: BSONDocument) -> Range<Int>? {
```
patrickfreed, line 257: Since this wasn't required by the benchmarks, I didn't originally update this to use the fast / unsafe `findValue`. It seems like a natural fit so I've updated it now.

Sources/SwiftBSON/BSONDocument.swift
patrickfreed, line 91: One thing that isn't reflected in this PR is that something like the following _could_ panic:
```swift
let doc = try BSONDocument(fromBSONWithoutValidatingElements: bson)
for i in 0..<doc.count {
    print(doc[i])
}
```
This is because the index-based subscript doesn't return an optional since it just `fatalErrors` when an out-of-bounds index is provided. Given that this subscript already can `fatalError` pretty easily, it seemed okay to me to `fatalError` on invalid BSON too, even if `count` says its okay. What do you think?

kmahar (approved):

PR #56, opened 2021-02-03
SWIFT-1086 Ignore key order in array-nested documents in equalsIgnoreKeyOrder

patrickfreed:
Fix for SWIFT-1086

### DIFF
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -570,14 +570,7 @@ extension BSONDocument {
         }
 
         for (k, v) in self {
-            let otherValue = other[k]
-            if case let (.document(docA), .document(docB)?) = (v, otherValue) {
-                guard docA.equalsIgnoreKeyOrder(docB) else {
-                    return false
-                }
-                continue
-            }
-            guard v == otherValue else {
+            guard let otherValue = other[k], v.equalsIgnoreKeyOrder(otherValue) else {
                 return false
             }
         }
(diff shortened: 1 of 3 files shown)


Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -570,14 +570,7 @@ extension BSONDocument {
...
         for (k, v) in self {
-            let otherValue = other[k]
-            if case let (.document(docA), .document(docB)?) = (v, otherValue) {
-                guard docA.equalsIgnoreKeyOrder(docB) else {
-                    return false
-                }
-                continue
-            }
-            guard v == otherValue else {
+            guard let otherValue = other[k], v.equalsIgnoreKeyOrder(otherValue) else {
```
kmahar, line 573: now that I think about it, shouldn't this method just not exist because a public `BSONDocument.equalsIgnoreKeyOrder` is already defined in the BSON library?

patrickfreed: Are you referring to `BSON.equalsIgnoreKeyOrder`? That's necessary (or rather helpful) to ease in recursion, since we have to go through `BSON` for both arrays and documents. If you mean `BSONDocument.sortedEquals` in the driver, then yeah that can be replaced with this. I originally tried doing that during the perf work but ran into this bug, so I had to implement it directly over there first. Once this is merged I'll open a follow-up PR updating the driver's implementation to just wrap this.

kmahar: lol whoops, I don't know why but I was thinking this was a PR against the driver itself 🤦‍♀️ nvm me

kmahar (approved):

PR #57, opened 2021-02-04
SWIFT-1088 ByteBuffer extJSON interoperability

patrickfreed:
SWIFT-1088

This PR adds APIs to `ExtendedJSONEncoder` and `ExtendedJSONDecoder` which utilize NIO's `ByteBuffer` instead of Foundation's `Data`. This can help reduce needless copies when using `ByteBuffer` that would otherwise be required when routing through `Data`.

### DIFF
--- a/Sources/SwiftBSON/ExtendedJSONDecoder.swift
+++ b/Sources/SwiftBSON/ExtendedJSONDecoder.swift
@@ -1,5 +1,6 @@
 import ExtrasJSON
 import Foundation
+import NIO
 
 /// `ExtendedJSONDecoder` facilitates the decoding of ExtendedJSON into `Decodable` values.
 public class ExtendedJSONDecoder {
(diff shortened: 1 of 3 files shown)


Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -65,6 +60,38 @@ public class ExtendedJSONDecoder {
...
+    public func decode<T: Decodable>(_: T.Type, from buffer: ByteBuffer) throws -> T {
+        guard buffer.readableBytes > 0 else {
+            throw DecodingError._extendedJSONError(keyPath: [], debugDescription: "empty buffer provided to decode")
+        }
+
+        var buffer = buffer
+        // readBytes never returns nil here because we checked that the buffer wasn't empty and only read
+        // readable bytes out from it.
+        // swiftlint:disable:next force_unwrapping
+        return try self.decodeBytes(T.self, from: buffer.readBytes(length: buffer.readableBytes)!)
```
patrickfreed, line 92: according to the `swift-extras-json` [readme](https://github.com/swift-extras/swift-extras-json#use-with-swiftnio-bytebuffer), it's more performant to read the bytes out of the buffer than it is to use `ByteBufferView` via `readableBytesView`. I think this is because the view relies on `readInteger` for each byte, which can be expensive to call repeatedly.

kmahar (approved):
LGTM! did get me thinking about whether we should be showing example usage of our encoders and decoders in READMEs either here or on the driver itself which I opened [SWIFT-1111](https://jira.mongodb.org/browse/SWIFT-1111) about (fun ticket number)

PR #61, opened 2021-02-10
SWIFT-1064 Rename ExtendedJSONEncoder.mode and make it a struct

kmahar:


Sources/SwiftBSON/ExtendedJSONEncoder.swift
```diff
@@ -29,14 +40,14 @@ public class ExtendedJSONEncoder {
...
         let encoder = BSONEncoder()
         encoder.userInfo = self.userInfo
         let bson: BSON = try encoder.encodeFragment(value)
 
         let json: JSON
-        switch self.mode {
+        switch self.format._format {
         case .canonical:
             json = bson.bsonValue.toCanonicalExtendedJSON()
         case .relaxed:
```
patrickfreed, line 53: the docstrings for `encode` and `encodeBuffer` have refereneces to `self.mode` that need to be updated.

patrickfreed (approved):
LGTM mod two docstring updates

Sources/SwiftBSON/ExtendedJSONEncoder.swift
kmahar: good catch, fixed 👍

PR #64, opened 2021-03-05
SWIFT-1126 / SWIFT-1137 Add legacy extended JSON parsing, fix Date-related crashes

patrickfreed:
SWIFT-1126 / SWIFT-1137

This PR adds support for legacy extended JSON v1 parsing, which the old libbson-based BSON library had support for. This PR also fixes potential crashes that may occur when trying to use `Date` values extremely far in the future (> `Int64.max` ms from epoch) or the past (< `Int64.min` ms before epoch).

### DIFF
--- a/Sources/SwiftBSON/BSONBinary.swift
+++ b/Sources/SwiftBSON/BSONBinary.swift
@@ -201,7 +201,7 @@ extension BSONBinary: BSONValue {
         // extended JSON v2
         case .object:
             guard obj.count == 1 else {
-                throw Swift.DecodingError.extraKeysError(
+                throw Swift.DecodingError._extraKeysError(
                     keyPath: keyPath,
                     expectedKeys: ["$binary"],
                     allKeys: Set(obj.keys)
@@ -239,7 +239,7 @@ extension BSONBinary: BSONValue {
             subtype = s
         case let .string(base64):
             guard obj.count == 2 else {
-                throw Swift.DecodingError.extraKeysError(
+                throw Swift.DecodingError._extraKeysError(
                     keyPath: keyPath,
                     expectedKeys: ["$binary"],
                     allKeys: Set(obj.keys)
--- a/Sources/SwiftBSON/BSONBinary.swift
+++ b/Sources/SwiftBSON/BSONBinary.swift
@@ -1,4 +1,5 @@
 import ExtrasBase64
+import ExtrasJSON
 import Foundation
 import NIO
 
@@ -57,6 +58,34 @@ public struct BSONBinary: Equatable, Hashable {
                 throw BSONError.InvalidArgumentError(message: "Cannot represent \(byteValue) as Subtype")
             }
             return subtype
+        }
+
+        fileprivate init(fromJSONValue json: JSONValue, keyPath: [String]) throws {
+            let subtypeInput: String
+            let errorMsg: String
+            switch json {
+            case let .string(s):
+                errorMsg = "subtype must be a BSON binary type as a one- or two-character hex string"
+                subtypeInput = s
+            case let .number(n):
+                errorMsg = "subtype must be a BSON binary type as an integer"
+                subtypeInput = n
+            default:
+                throw Swift.DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "subtype must be a BSON binary type, got \(json) instead"
+                )
+            }
+            guard
+                let subTypeInt = UInt8(subtypeInput, radix: 16),
+                let subType = Subtype(rawValue: subTypeInt)
+            else {
+                throw Swift.DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `SubType` from \"\(json)\", \(errorMsg)"
+                )
+            }
+            self = subType
         }
     }
 
@@ -146,6 +175,7 @@ public struct BSONBinary: Equatable, Hashable {
 
 extension BSONBinary: BSONValue {
     internal static let extJSONTypeWrapperKeys: [String] = ["$binary", "$uuid"]
+    internal static let extJSONLegacyTypeWrapperKeys: [String] = ["$type"]
 
     /*
      * Initializes a `Binary` from ExtendedJSON.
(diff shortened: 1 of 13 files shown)


Sources/SwiftBSON/BSONBinary.swift
```diff
@@ -188,38 +190,104 @@ extension BSONBinary: BSONValue {
...
+                throw Swift.DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `SubType` from \"\(json)\", subtype must"
+                        + "be a BSON binary type as a one- or two-character hex string"
+                )
+            }
+
+            base64Str = b64Str
+            subtype = s
+        case let .string(base64):
```
patrickfreed, line 240: this parses the following format:
```
{
    "$binary": <base64 string>,
    "$type": <hex string or number>
}
```

Sources/SwiftBSON/ExtendedJSONDecoder.swift
```diff
@@ -17,8 +17,15 @@ public class ExtendedJSONDecoder {
     }()
 
     /// A set of all the possible extendedJSON wrapper keys.
+    /// This does not include the legacy extended JSON wrapper keys.
```
patrickfreed, line 20: Since we allow things like "$type" to be in a regular document (e.g. as part of a query operator), the legacy wrapper keys can't be used to determine errors like we do with the reserved ones, so we need to separate them out.

Tests/SwiftBSONTests/CommonTestUtils.swift
```diff
@@ -62,6 +62,25 @@ public func sortedEqual(_ expectedValue: BSONDocument?) -> Predicate<BSONDocumen
     }
 }
 
+public func sortedEqual(_ expectedValue: BSON?) -> Predicate<BSON> {
```
patrickfreed, line 65: This is just a copy/paste that extends this matcher to `BSON`

Tests/SwiftBSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -542,4 +539,185 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
+        Self.jsonTest(json: invalidStringSubtype, expectation: .error())
+
+        let invalidStringNumberSubtype: JSON = [
+            "$binary": JSON(.string(base64)),
+            "$type": "12345"
+        ]
+        Self.jsonTest(json: invalidStringNumberSubtype, expectation: .error())
+    }
+
+    func testLegacyExtendedJSONDate() throws {
```
patrickfreed, line 659: these test cases are taken from libbson

Tests/SwiftBSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -542,4 +539,185 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
+    func testLegacyExtendedJSONRegex() throws {
+        let regex = BSONRegularExpression(pattern: "abc", options: "ix")
+
+        Self.jsonTest(
+            json: ["val": ["$regex": "abc", "$options": "ix"]],
+            expectation: .success(regex)
+        )
+
+        // don't invalidate a "$regex" query operator stored in JSON
+        Self.jsonTest(
```
patrickfreed, line 706: the first of these 3 were taken from libbson too, except libbson would return an error. That seemed problematic to me, since these could easily be valid `$regex` query operators, just stored in JSON. I figured no longer throwing an error in these cases would not really constitute a breaking change, and that the legacy parsing should defer to the non-legacy path if it ever is ambiguous.

Sources/SwiftBSON/BSONRegularExpression.swift
```diff
@@ -81,22 +82,36 @@ extension BSONRegularExpression: BSONValue {
...
+                throw DecodingError._extendedJSONError(
+                    keyPath: keyPath,
+                    debugDescription: "Could not parse `BSONRegularExpression` from \"\(regex)\", " +
+                        "\"pattern\" and \"options\" must be strings"
+                )
+            }
+            self = BSONRegularExpression(pattern: patternStr, options: optionsStr)
+            return
+        } else {
+            // legacy / v1 extended JSON
```
patrickfreed, line 101: this parses the following format:
```
{
    "$regex": <string>,
    "$options": <string>,
}
```
which is unfortunately, the exact same format as a [query operator](https://docs.mongodb.com/manual/reference/operator/query/regex/).

Sources/SwiftBSON/Date+BSONValue.swift
```diff
@@ -48,6 +51,15 @@ extension Date: BSONValue {
                 )
             }
             self = date
+        case let .number(ms):
```
patrickfreed, line 54: legacy allows for a regular number instead of a `$numberLong` wrapper.

Sources/SwiftBSON/Date+BSONValue.swift
```diff
@@ -87,7 +99,20 @@ extension Date: BSONValue {
     internal var bson: BSON { .datetime(self) }
 
     /// The number of milliseconds after the Unix epoch that this `Date` occurs.
-    internal var msSinceEpoch: Int64 { Int64((self.timeIntervalSince1970 * 1000.0).rounded()) }
+    /// If the date is further in the future than Int64.max milliseconds from the epoch,
+    /// Int64.max is returned to prevent a crash.
+    internal var msSinceEpoch: Int64 {
```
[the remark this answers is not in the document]
patrickfreed: good idea, done

Tests/SwiftBSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -542,4 +539,185 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
...
+            json: ["val": ["$date": JSON(.number("-10000000000000000000"))]],
+            expectation: .error()
+        )
+    }
+
+    func testLegacyExtendedJSONRegex() throws {
+        let regex = BSONRegularExpression(pattern: "abc", options: "ix")
+
+        Self.jsonTest(
+            json: ["val": ["$regex": "abc", "$options": "ix"]],
```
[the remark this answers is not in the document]
patrickfreed: It's not ideal, but this is the existing behavior in libbson, so I think we need to preserve it.

kmahar (approved):
lgtm

PR #67, opened 2021-07-30
SWIFT-1107, SWIFT-964: BSON corpus test updates

kmahar:
* Sync BSON corpus test changes made as part of the "Mitigate pain of using field names with dots and dollars" project (SWIFT-1107)
* I noticed that SWIFT-963 seems to be "gone away" as the tests all pass; so added SWIFT-964 assertions to round trip degenerate BSON through native types

### DIFF
--- a/Tests/SwiftBSONTests/BSONCorpusTests.swift
+++ b/Tests/SwiftBSONTests/BSONCorpusTests.swift
@@ -192,9 +192,10 @@ final class BSONCorpusTests: BSONTestCase {
 
                         let docFromDB = try BSONDocument(fromBSON: dBData)
 
-                        // SKIPPING: native_to_bson( bson_to_native(dB) ) = cB
-                        // We only validate the BSON bytes, we do not clean them up, so can't do this assertion
-                        // Degenerate BSON round trip tests will be added in SWIFT-964
+                        // native_to_bson( bson_to_native(dB) ) = cB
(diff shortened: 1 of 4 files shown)


Tests/SwiftBSONTests/BSONCorpusTests.swift
```diff
@@ -199,9 +192,10 @@ final class BSONCorpusTests: BSONTestCase {
 
                         let docFromDB = try BSONDocument(fromBSON: dBData)
 
-                        // SKIPPING: native_to_bson( bson_to_native(dB) ) = cB
-                        // We only validate the BSON bytes, we do not clean them up, so can't do this assertion
```
kmahar, line 203: I don't think this comment is quite accurate as we can still test we have the right behavior (as I do here) without cleaning up the bytes on document initialization.

I guess we could have written an assertion like
```swift
expect(BSONDocument(fromBSON: dBData).toData()).to(equal(cB))
```
but I'm not actually sure if modifying the degenerate bytes is expected/desired behavior. it seems like it could potentially mess up round trip-ability with the server. what do you think?

patrickfreed: I think what we have seems to be the best of both worlds: we only yield correct values but we also preserve round trip-ability, so your assertion LGTM.

patrickfreed (approved):

PR #68, opened 2021-09-08
SWIFT-871, SWIFT-1340: Evergreen config updates

kmahar:
* Test on Ubuntu 20.04 rather than 16.04
* Test on macOS 10.15 rather than 10.14
* Add testing on Swift 5.4 and 5.5-dev
* Add TSan task
* bump Swift version used for format/lint and bump SwiftLint version accordingly

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -102,6 +102,7 @@ functions:
         script: |
           ${PREPARE_SHELL}
           SWIFT_VERSION=${SWIFT_VERSION} \
+          SANITIZE=${SANITIZE} \
           ${PROJECT_DIRECTORY}/.evergreen/run-tests.sh
   
   "upload test results":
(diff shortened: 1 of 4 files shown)


kmahar:
As an aside, I'm not sure how valuable TSan really is in the context of the BSON library as compared to the driver, but figured we may as well have it on 🤷‍♀️

.evergreen/config.yml
```diff
@@ -157,37 +158,69 @@ axes:
...
+        - "5.3"
+        - "5.4"
+        - "5.5-dev"
     tasks:
       - name: "test"
 
-# define a separate matrix for 5.3 - no way to remove 5.3 + macOS from the definition above.
-# see EVG-13092
-  - matrix_name: "tests-all-5.3"
+  - matrix_name: "tests-5.1"
```
patrickfreed, line 198: do we want to do the compile-only setup on 5.1 here too?

kmahar: oh, I meant to add a comment about this. I thought about this, but the entire suite of tests for this library only take 2.5s seconds to run (for me locally anyway), so the savings are pretty minimal and I wasn't sure it was worth adding the complexity to our Evergreen setup.

patrickfreed: Ah okay, sounds good.

patrickfreed (approved):

kmahar:
going to hold off on merging pending investigation into using `leaks` for leak checking on macOS since we'd probably want to incorporate that here too

kmahar:
Added a leak checking task.
[successful](https://spruce.mongodb.com/version/6140e0bf57e85a28319b6e10/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[intentional leak](https://spruce.mongodb.com/version/6140e150e3c33143ed2a02d2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

patrickfreed (approved):

PR #69, opened 2021-09-14
Decouple BSONEncoder internals from BSONValue protocol

kmahar:
This is the way this code should have been in the first place, I think 🙂

Previously, we used the `BSONValue` protocol as the intermediate representation of data types stored in the encoder before the final conversion to BSON. However, the only requirement we actually needed from the protocol was the `.bson` property. 

As a result, our internal data structures `MutableDictionary` and `MutableArray` had to conform to `BSONValue`.

This refactor introduces a new `BSONRepresentable` profile that is used throughout the encoder in place of `BSONValue` that actually expresses what we need for encoding purposes - the ability for a type to be converted to BSON.

This is a precursor to work to fix the null bytes issue, will expand in comments.

### DIFF
--- a/Sources/SwiftBSON/BSONEncoder.swift
+++ b/Sources/SwiftBSON/BSONEncoder.swift
@@ -760,18 +760,19 @@ extension _BSONEncoder: SingleValueEncodingContainer {
     }
 }
 
+/// Indicates that a type can be converted to a `BSON`.
 internal protocol BSONRepresentable {
     func toBSON() throws -> BSON
 }
 
 extension Array: BSONRepresentable where Element == BSON {
-    internal func toBSON() throws -> BSON {
+    internal func toBSON() -> BSON {
         .array(self)
     }
 }
 
 extension BSONValue {
-    internal func toBSON() throws -> BSON {
+    internal func toBSON() -> BSON {
         self.bson
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/SwiftBSON/BSONEncoder.swift
```diff
@@ -760,79 +760,61 @@ extension _BSONEncoder: SingleValueEncodingContainer {
     }
 }
 
+internal protocol BSONRepresentable {
+    func toBSON() throws -> BSON
```
kmahar, line 765: this is marked as throws (even though all current implementations do not require throwing), because it will need to start throwing as part of the null bytes fix. (see below)

Sources/SwiftBSON/BSONEncoder.swift
```diff
@@ -859,46 +841,20 @@ private class MutableDictionary: BSONValue {
     }
 
     /// Converts self to a `BSONDocument` with equivalent key-value pairs.
-    fileprivate func toDocument() -> BSONDocument {
+    fileprivate func toDocument() throws -> BSONDocument {
         var doc = BSONDocument()
         for i in 0..<self.keys.count {
             let value = self.values[i]
-            doc.append(key: self.keys[i], value: value.bson)
+            doc.append(key: self.keys[i], value: try value.toBSON())
```
kmahar, line 849: `doc.append` will start throwing as part of the null bytes fix, because it calls through to `ByteBuffer.writeCString` which will start throwing as well.
altogether, this will allow us to error if a user tries to encode something containing a null byte in one of the coding keys, rather than crashing.

patrickfreed (approved):
LGTM!

PR #70, opened 2021-09-27
SWIFT-1260 Disallow embedded null bytes in BSON keys

kmahar:
I did some more re-running of the benchmarks and got the following results:

* JSON to BSON: flat ~5% slower; deep ~7% slower; full  ~10% slower, multi-import ~8% slower
* native to BSON: flat no change, deep ~10% slower, full no change
* runCommand, findOne, insertOne varied across runs, occasionally getting both faster and slower with differences in the range of 1-3%...

For some other benchmarks which do not even touch the new `isValidCString` code, I actually saw both slight speedups and slight slowdowns (like <= 5% change)... I think those can probably just be chalked up to inconsistency on the spawn hosts. 

These numbers do not sound great, but revisiting the scope for the recent BSON performance project, compared to the scope of the performance hits we accepted there (300% slower for JSON to BSON, 25% slower for native to BSON) I don't think these are catastrophic. At the end of the day there is no way around doing this validation so I think we ought to go ahead as-is and if perf comes up as an issue for our users we can revisit and look for other ways to speed up the library.

In summary, what has changed here:
* Synced BSON corpus tests and added new prose tests
* The method we use to encode a C string to a `ByteBuffer`, `writeCString`, now validates there are no embedded null bytes before writing
* As a result of the previous bullet, a number of internal methods now `throw`, hence the number of new `try`s inserted throughout the code
* As a result of the previous bullet, we have to do some new do/catch-ing in places we call those internal methods but are not allowed to throw
* When initializing a `BSONRegularExpression` from JSON, we validate that the pattern and options are valid C strings

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/Quick/Nimble.git",
         "state": {
           "branch": null,
-          "revision": "2b1809051b4a65c1d7f5233331daa24572cd7fca",
-          "version": "8.1.1"
+          "revision": "7a46a5fc86cb917f69e3daf79fcb045283d8f008",
+          "version": "8.1.2"
         }
       },
       {
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio",
         "state": {
           "branch": null,
-          "revision": "120acb15c39aa3217e9888e515de160378fbcc1e",
-          "version": "2.18.0"
+          "revision": "6aa9347d9bc5bbfe6a84983aec955c17ffea96ef",
+          "version": "2.33.0"
         }
       }
     ]
--- a/Sources/SwiftBSON/BSON.swift
+++ b/Sources/SwiftBSON/BSON.swift
@@ -450,7 +450,11 @@ extension BSON: ExpressibleByBooleanLiteral {
 
 extension BSON: ExpressibleByDictionaryLiteral {
     public init(dictionaryLiteral elements: (String, BSON)...) {
-        self = .document(BSONDocument(dictionaryLiteral: elements))
+        do {
+            self = .document(try BSONDocument(dictionaryLiteral: elements))
+        } catch {
+            fatalError("\(error)")
+        }
     }
 }
 
--- a/Sources/SwiftBSON/BSONDocument.swift
+++ b/Sources/SwiftBSON/BSONDocument.swift
@@ -17,11 +17,7 @@ public struct BSONDocument {
 
     internal var storage: BSONDocumentStorage
 
-    internal init(_ elements: [BSON]) {
-        self = BSONDocument(keyValuePairs: elements.enumerated().map { i, element in (String(i), element) })
-    }
-
-    internal init(keyValuePairs: [(String, BSON)]) {
+    internal init(keyValuePairs: [(String, BSON)]) throws {
         self.storage = BSONDocumentStorage()
 
         let start = self.storage.buffer.writerIndex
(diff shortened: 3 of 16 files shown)


Sources/SwiftBSON/BSON.swift
```diff
@@ -450,7 +450,11 @@ extension BSON: ExpressibleByBooleanLiteral {
 
 extension BSON: ExpressibleByDictionaryLiteral {
     public init(dictionaryLiteral elements: (String, BSON)...) {
-        self = .document(BSONDocument(dictionaryLiteral: elements))
+        do {
```
kmahar, line 453: the `ExpressibleByDictionaryLiteral` protocol does not allow us to throw from this initializer, so we must crash if the input dictionary literal is invalid

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -17,11 +17,7 @@ public struct BSONDocument {
 
     internal var storage: BSONDocumentStorage
 
-    internal init(_ elements: [BSON]) {
```
kmahar, line 20: this initializer was not being used anywhere anymore so I just got rid of it.

Sources/SwiftBSON/BSONDocument.swift
```diff
@@ -185,7 +181,12 @@ public struct BSONDocument {
             BSONDocumentIterator.find(key: key, in: self)?.value
         }
         set {
-            self.set(key: key, to: newValue)
+            // We have no choice but to crash here.
```
kmahar, line 184: Similarly, we are not allowed to throw from subscript setters/getters. (well, we can as of Swift 5.5, but we support Swift 5.1+ right now). I opened SWIFT-1357 about starting to throw from the subscript in our next major version

Package.resolved
```diff
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio",
         "state": {
           "branch": null,
-          "revision": "120acb15c39aa3217e9888e515de160378fbcc1e",
-          "version": "2.18.0"
+          "revision": "6aa9347d9bc5bbfe6a84983aec955c17ffea96ef",
+          "version": "2.33.0"
```
kmahar, line 37: not strictly related to this PR but NIO required an update to build on Swift 5.5

patrickfreed (commented):
one suggestion for a test assertion but otherwise LGTM

Tests/SwiftBSONTests/BSONCorpusTests.swift
```diff
@@ -285,4 +291,89 @@ final class BSONCorpusTests: BSONTestCase {
             }
         }
     }
+
+    /// Tests from the "Prose Tests" section of the test plan.
+    func testBSONCorpusProse() throws {
+        // 1. Prohibit null bytes in null-terminated strings when encoding BSON
+        // Note we can't write these tests using the dictionary literal syntax since all of these cases trigger
+        // fatalErrors with that API.
```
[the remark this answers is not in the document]
kmahar: nice, I added additional expectations that test the public API explicitly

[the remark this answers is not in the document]
patrickfreed: Unfortunately it looks like using that in combination with the leak checker crashes the compiler per the evergreen output. Maybe we just have to go back to the original test after all :disappointed:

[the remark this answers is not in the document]
kmahar: ugh. yeah I guess so. will revert my last two commits

patrickfreed (approved):
LGTM!

isabelatkinson (approved):

PR #74, opened 2021-11-29
Add BSON Binary Subtype 7

nspektor:


.swift-version
```diff
@@ -1 +1 @@
-5.2.4
+5.5.1
```
nspektor, line 1: Should this change be part of the PR? or should I somehow only do it locally? (It was required for tests to pass)

nspektor (commented):
Is there anything else that needs to be done to support this new column subtype?

.swift-version
isabelatkinson: hm I'm not sure if we want to bump the version here, @kmahar or @patrickfreed could probably give a better answer

kmahar: yeah I think it's fine to update this now

kmahar (approved):
lgtm!

isabelatkinson (approved):
lgtm!

.swift-version
nspektor: Great, thanks

PR #78, opened 2022-02-16
Make necessary changes to build and test the library on iOS

kmahar:
These changes allow the library to build and tested on iOS, either directly in Xcode or from the command line using `xcodebuild`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,8 @@ import PackageDescription
 let package = Package(
     name: "swift-bson",
     platforms: [
-        .macOS(.v10_14)
+        .macOS(.v10_14),
+        .iOS(.v11)
     ],
     products: [
         .library(name: "SwiftBSON", targets: ["SwiftBSON"])
--- a/Tests/SwiftBSONTests/BSONTests.swift
+++ b/Tests/SwiftBSONTests/BSONTests.swift
@@ -6,19 +6,15 @@ import SwiftBSON
 import XCTest
 
 open class BSONTestCase: XCTestCase {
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -4,7 +4,8 @@ import PackageDescription
 let package = Package(
     name: "swift-bson",
     platforms: [
-        .macOS(.v10_14)
+        .macOS(.v10_14),
+        .iOS(.v11)
```
kmahar, line 8: we use some Swift APIs that are only available on iOS 11+, so this is the lowest possible value of this. But per the scope I need to do a little research to figure out what a reasonable choice for this is before we release.

Tests/SwiftBSONTests/BSONTests.swift
```diff
@@ -6,19 +6,15 @@ import SwiftBSON
...
-        // otherwise we're in Xcode, get the bundle's resource path
-        guard let path = Bundle(for: self).resourcePath else {
-            XCTFail("Missing resource path")
-            return ""
-        }
-        return path
+        // Approach taken from https://stackoverflow.com/a/58034307
+        // TODO: SWIFT-1442 Once we drop Swift < 5.3 we can switch to including the JSON files as Resources via our
+        // package manifest instead.
+        let thisFile = URL(fileURLWithPath: #file)
```
isabelatkinson, line 14: what is `#file` here?

kmahar: it gives the full path to the file it appears in - so in this case it would give the full path to `Tests/SwiftBSONTests/BSONTests.swift`.

PR #79, opened 2022-02-22
Update tools versions

kmahar:
Updates the versions of SwiftLint and SwiftFormat used in CI

### DIFF
--- a/.evergreen/install-tools.sh
+++ b/.evergreen/install-tools.sh
@@ -39,10 +39,10 @@ swiftenv local $SWIFT_VERSION
 
 if [ $1 == "swiftlint" ]
 then
-    build_from_gh swiftlint https://github.com/realm/SwiftLint "0.44.0"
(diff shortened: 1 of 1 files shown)


.evergreen/install-tools.sh
```diff
@@ -39,10 +39,10 @@ swiftenv local $SWIFT_VERSION
 
 if [ $1 == "swiftlint" ]
 then
-    build_from_gh swiftlint https://github.com/realm/SwiftLint "0.44.0"
+    build_from_gh swiftlint https://github.com/realm/SwiftLint "0.46.2"
```
isabelatkinson, line 42: looks like `0.46.3` was just [released](https://github.com/realm/SwiftLint/releases/tag/0.46.3) yesterday, do we want to use that version instead? (ditto other PR)

kmahar: good catch, I was just looking at the latest version that homebrew had which lags a bit. updated

isabelatkinson (approved):

PR #80, opened 2022-03-07
SWIFT-1500, SWIFT-1501: Set minimum iOS version and expand iOS testing

kmahar:
Sets the minimum iOS version for the library to iOS 13, and updates our GH Actions workflow to test on each supported iOS version.

The updated workflow file here applies to what was run for this PR, so you can see the generated tasks that way.

To choose the exact Xcode versions I referenced this [documentation](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#installed-sdks) which shows which iOS simulators are available with each Xcode version, and just went with the latest Xcode + latest iOS patch for each of iOS 13, 14, and 15. 

This implicitly has us test with the corresponding Swift versions that shipped with each of those Xcode versions, which is 5.5.2, 5.4.2, and 5.2.4 respectively. 

From a brief survey of other libraries supporting iOS (Realm, RxSwift, AlamoFire, Nimble), it doesn't seem to be common practice to try to test multiple Swift versions across single iOS versions, so for now I figured this would probably be sufficient coverage and we didn't need to bother with e.g. trying to test every Swift version the library supports across every iOS version we support.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build-and-test-iOS:
-    name: Build and Test on iOS
+    name: Build and Test on iOS ${{ matrix.iOS-version }}
     runs-on: macos-11
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/ios.yml
```diff
@@ -7,16 +7,32 @@ on:
...
 
 jobs:
-  build:
-    name: Build and Test on iOS
-    runs-on: macos-latest
+  build-and-test-iOS:
+    name: Build and Test on iOS ${{ matrix.iOS-version }}
+    runs-on: macos-11
+    strategy:
+      fail-fast: false
```
kmahar, line 14: with fail-fast on, if one generated task fails the rest will all be aborted. this didn't seem very useful since it would make it hard to tell if something was only failing on one iOS version or all.

isabelatkinson (approved):
lgtm, the testing matrix you chose seems reasonable to me

patrickfreed (approved):
LGTM! Awesome to see how simple the config is to get this running

PR #83, opened 2022-06-14
SWIFT-1508 Conform BSONTimestamp to Comparable

rchhaya:
Adds adherence of the `BSONTimestamp` type to the `Comparable` Protocol by implementing the < and == operator functions as defined [here](https://developer.apple.com/documentation/swift/comparable). Also modifies the `testTimestamp()` case to compare two instances of `BSONTimestamp` in addition to extended JSON testing

### DIFF
--- a/Sources/SwiftBSON/BSONTimestamp.swift
+++ b/Sources/SwiftBSON/BSONTimestamp.swift
@@ -91,9 +91,8 @@ public struct BSONTimestamp: BSONValue, Comparable, Equatable, Hashable {
         ]
     }
 
-    /// Compares two `BSONTimestamp` values as outlined by the Comparable protocol
+    /// Compares two `BSONTimestamp` instances as outlined by the Comparable protocol
     public static func < (lhs: BSONTimestamp, rhs: BSONTimestamp) -> Bool {
-        // If ts same move onto comparabel
         if lhs.timestamp != rhs.timestamp {
             return lhs.timestamp < rhs.timestamp
         } else { // equal timestamps, look at increment
@@ -101,9 +100,9 @@ public struct BSONTimestamp: BSONValue, Comparable, Equatable, Hashable {
         }
     }
 
-    /// Checks two `BSONTimestamp`s for equality as outlined by the Comparable protocol
+    /// Checks two `BSONTimestamp` instances for equality as outlined by the Comparable protocol
     public static func == (lhs: BSONTimestamp, rhs: BSONTimestamp) -> Bool {
-        lhs.timestamp == rhs.timestamp && lhs.increment == rhs.increment
+        return lhs.timestamp == rhs.timestamp && lhs.increment == rhs.increment
     }
 
     internal static func read(from buffer: inout ByteBuffer) throws -> BSON {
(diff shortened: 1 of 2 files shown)


Sources/SwiftBSON/BSONTimestamp.swift
```diff
@@ -91,6 +91,20 @@ public struct BSONTimestamp: BSONValue, Equatable, Hashable {
         ]
     }
 
+    /// Compares two `BSONTimestamp` instances as outlined by the Comparable protocol
```
kmahar, line 94:
```suggestion
    /// Compares two `BSONTimestamp` instances as outlined by the `Comparable` protocol.
```

(generally we end sentences in docstrings with periods)

Sources/SwiftBSON/BSONTimestamp.swift
```diff
@@ -91,6 +91,20 @@ public struct BSONTimestamp: BSONValue, Equatable, Hashable {
...
+    public static func < (lhs: BSONTimestamp, rhs: BSONTimestamp) -> Bool {
+        if lhs.timestamp != rhs.timestamp {
+            return lhs.timestamp < rhs.timestamp
+        } else { // equal timestamps, look at increment
+            return lhs.increment < rhs.increment
+        }
+    }
+
+    /// Checks two `BSONTimestamp` instances for equality as outlined by the Comparable protocol
```
kmahar, line 103:
```suggestion
    /// Checks two `BSONTimestamp` instances for equality as outlined by the `Comparable` protocol.
```

Tests/SwiftBSONTests/ExtendedJSONConversionTests.swift
```diff
@@ -373,6 +373,15 @@ open class ExtendedJSONConversionTestCase: BSONTestCase {
             .to(throwError(errorType: DecodingError.self))
         expect(try BSONTimestamp(fromExtJSON: ["$timestamp": ["t": 1, "i": 2], "extra": "2"], keyPath: []))
             .to(throwError(errorType: DecodingError.self))
+
+        // Comparable testing
```
kmahar, line 377: this test file, and all the tests it currently contains, are specifically about converting BSON types to and from extended JSON. I think it would be better to pull this code out into its own test case. maybe in `BSONValueTests.swift`?

kmahar (changes requested):
looks good! a few minor comments.

Sources/SwiftBSON/BSONTimestamp.swift
```diff
@@ -91,6 +91,20 @@ public struct BSONTimestamp: BSONValue, Equatable, Hashable {
...
+    public static func < (lhs: BSONTimestamp, rhs: BSONTimestamp) -> Bool {
+        if lhs.timestamp != rhs.timestamp {
+            return lhs.timestamp < rhs.timestamp
+        } else { // equal timestamps, look at increment
+            return lhs.increment < rhs.increment
+        }
+    }
+
+    /// Checks two `BSONTimestamp` instances for equality as outlined by the `Comparable` protocol.
+    public static func == (lhs: BSONTimestamp, rhs: BSONTimestamp) -> Bool {
```
patrickfreed, line 104: We actually don't need to supply this custom implementation, since the auto generated one behaves the same way (this is what we were using before actually via the `Equatable` conformance).

We do need to keep the custom `Comparable` conformance in order to specify the precedence between timestamp and increment though.

patrickfreed (approved):

isabelatkinson (approved):

PR #85, opened 2022-06-20
swift-bson doesn't seem to have to rely on all modules in NIO

oozoofrog:
- NIO dependency change to NIOCore
- import NIO change to import NIOCore


kmahar:
Hi @oozoofrog, thanks for the PR and bringing this to our attention! 

You're right that the library is only relying on types from NIOCore. At the time we wrote the library, SwiftNIO had not been broken up into the smaller modules yet. This seems like a very nice improvement -- from my rough testing this change shaves ~20% off compilation time and ~30% off the size of a Swift executable that depends only on SwiftBSON. 😎

One slightly tricky thing about this change is that the SwiftNIO version where NIOCore was introduced only supports Swift 5.2 and up, and this library currently supports Swift 5.1+. We could do some `#if`/`#else` checks around the import statements to work around that, but for simplicity we've decided we'll go ahead and drop Swift 5.1 altogether in our next minor version, 3.2.0. I've scheduled [SWIFT-1337](https://jira.mongodb.org/browse/SWIFT-1337) which tracks the necessary work for that.

To avoid breaking CI I'm going to hold off on merging this until we've completed SWIFT-1337.

Thanks for the contribution!

kmahar (approved):

kmahar:
@patrickfreed, @isabelatkinson this is ready for a look now!

isabelatkinson (approved):

nbbeeken (approved):

patrickfreed (approved):

kmahar:
@oozoofrog thanks for the contribution!

oozoofrog:
> @oozoofrog thanks for the contribution!

thanks for merge 😄

PR #88, opened 2022-10-20
SWIFT-1660 publicize [UInt8]- and ByteBuffer-taking initializers on BSONBinary

tayloraswift:
publicizes the `[UInt8]`- and `ByteBuffer`-taking initializers on `BSONBinary`. fixes [SWIFT-1660](https://jira.mongodb.org/browse/SWIFT-1660).

### DIFF
--- a/Sources/SwiftBSON/BSONBinary.swift
+++ b/Sources/SwiftBSON/BSONBinary.swift
@@ -89,12 +89,20 @@ public struct BSONBinary: Equatable, Hashable {
         self = try BSONBinary(buffer: buffer, subtype: subtype)
     }
 
+    /// Initializes a `BSONBinary` instance from a `[UInt8]` array and a `Subtype` subtype.
+    /// This will always create a copy of the array contents.
+    /// - Throws:
+    ///   - `BSONError.InvalidArgumentError` if the provided array is incompatible with the specified subtype.
     public init(bytes: [UInt8], subtype: Subtype) throws {
         var buffer = BSON_ALLOCATOR.buffer(capacity: bytes.count)
         buffer.writeBytes(bytes)
         self = try BSONBinary(buffer: buffer, subtype: subtype)
     }
 
+    /// Initializes a `BSONBinary` instance from a `ByteBuffer` and a `Subtype` subtype.
+    /// This will always create a copy of the buffer contents.
+    /// - Throws:
+    ///   - `BSONError.InvalidArgumentError` if the provided buffer is incompatible with the specified subtype.
     public init(buffer: ByteBuffer, subtype: Subtype) throws {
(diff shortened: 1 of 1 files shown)


Sources/SwiftBSON/BSONBinary.swift
```diff
@@ -89,13 +89,13 @@ public struct BSONBinary: Equatable, Hashable {
         self = try BSONBinary(buffer: buffer, subtype: subtype)
     }
 
-    internal init(bytes: [UInt8], subtype: Subtype) throws {
+    public init(bytes: [UInt8], subtype: Subtype) throws {
```
kmahar, line 92: mind adding some basic docstrings to these? I think they would be pretty similar to the ones for the init that takes a `Data` (same error case re: incompatible subtype).

kmahar (commented):
thanks for the contribution!

Sources/SwiftBSON/BSONBinary.swift
kelvin13: done

Sources/SwiftBSON/BSONBinary.swift
```diff
@@ -89,13 +89,21 @@ public struct BSONBinary: Equatable, Hashable {
...
+    ///   - `BSONError.InvalidArgumentError` if the provided array is incompatible with the specified subtype.
+    public init(bytes: [UInt8], subtype: Subtype) throws {
         var buffer = BSON_ALLOCATOR.buffer(capacity: bytes.count)
         buffer.writeBytes(bytes)
         self = try BSONBinary(buffer: buffer, subtype: subtype)
     }
 
-    internal init(buffer: ByteBuffer, subtype: Subtype) throws {
+    /// Initializes a `BSONBinary` instance from a `ByteBuffer` and a `Subtype` subtype.
+    /// This will always create a copy of the buffer contents.
```
kmahar, line 103: I'm not sure if this line is true in this case, because I believe `ByteBuffer` implements copy-on-write semantics (see various uses of `isKnownUniquelyReferenced` here: https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/ByteBuffer-core.swift) 
so maybe we should just remove this line?

kmahar (commented):
lgtm mod one small comment

patrickfreed (approved):
Looks good mod kaitlins comment about CoW, thanks for the PR!

Sources/SwiftBSON/BSONBinary.swift
kelvin13: ah great catch, i had just copied that sentence over blindly

kmahar (approved):
LGTM! we're planning to do both BSON and driver releases sometime in the next few weeks so this will go out with those

